PR #19, opened 2023-06-23
[F] Authentication POC

dananjohnson:
## What this change does ##

This is a proof of concept for authenticating visitors against user records in the Craft database. Similar to the approach on the main Rubin site, it integrates with the Craft [GraphQL Authentication plugin](https://graphql-authentication.jamesedmonston.co.uk/) to send authentication requests as mutations to Craft and receive tokens back. The key difference and the reason for doing this POC is that this approach tries to integrate with Server Components and server-side rendering, rather than strictly holding auth state and methods in the client with React hooks and the Context API. The goal of trying this alternative approach was to come up with a solution to authentication/authorization that fit better with React 18 and Next.js 13 and their recent emphasis on the server as the default place to do things. In addition, I wanted to utilize TypeScript to make managing auth and working with the API feel more predictable.

The general concept is as follows: when someone registers or signs in on an investigation landing page, a request is made through the server using a [Server Action](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions). If successful, the API will return a JWT and refresh token (plus expiration values) that are then stored as cookies. They will also be redirected to the first child route of the investigation (to be defined later, probably as a property of the investigation in Craft). Child routes are all authorized: when requested, the server will check cookies for the user's JWT, and if not found, will send them back to the landing page to sign back in.

This PR does not worry about certain nuances like handling pending status, or student vs. educator roles, or refreshing expired JWTs. But it does show that as a general design, this approach should work. The next step would be to implement these particular details, as well as to bring the UI/UX more in-line with what appears in design.

## Notes for reviewers ##

If you want to test this, you'll need to be running the `kp/gql-auth` branch in the API repo.

Once running, you should be able to confirm a few things:

1. You can register as a new user by visiting any child route of `localhost:3000` (e.g. `localhost:3000/some-investigation`; corresponding to `[locale]/[investigation]` in the app router).
3. You can sign in as that user (status needs to be updated to active in the backend currently).
4. You can access an authorized child route of an investigation when signed in.
5. You can sign out, which will redirect you back to the investigation landing page and empty your auth cookies.

5: this is just a proof of concept, so it doesn't need a super detailed review. I'm more interested in 👍🏻 or 👎🏻 on the general approach, and if you have any specific concerns.

## Checklist ##

- [X] This change impacts several investigations (e.g. the change affects reuseed styles, widgets, pages, QAs, utility methods, etc.)


alexgoff (approved):
checked through and this POC looks good to me, I appreciate sticking to the structure of app router and taking advantage of it to move authentication to the server. Good to move ahead!

dananjohnson:
Great, thanks for the review @alexgoff! I'll go ahead and merge this once the corresponding API repo PR is merged, then will open up a new PR against `main` for additional auth work.

PR #36, opened 2023-08-07
[F] Finish implementing Facebook SSO

### LINKED ISSUE #25: Finalize authentication in client

dananjohnson:
Resolves #25 

## What this change does ##

This PR adds the missing dialog which sends a mutation to the API to register and authenticate a user signing in with their Facebook account.

## Notes for reviewers ##

Note that this feature won't fully work if the Facebook app isn't successfully authenticating you and redirecting you back to `localhost`. This was the case for me, but after going back and forth with Blake it seems like the issue is with my local or with me not being an admin on the Facebook app. He was able to go through this process just fine. See the thread in Slack, near the bottom, for the discussion.

## Testing ##

Make sure to update your API repo to use `http://localhost:3000/?facebook_sso=true` for the Facebook redirect URL.

## Checklist ##

- [X] This change is isolated to a specific page or Component (i.e. it's a one-off)


components/auth/dialogs/SignInFacebook/SignInFacebook.tsx
```diff
@@ -0,0 +1,100 @@
...
+        return t("facebook_sso.loading");
+      default:
+        return undefined;
+    }
+  }
+
+  function getDescription() {
+    switch (status) {
+      case "success":
+        return t("facebook_sso.success_message", { context: pendingGroup });
```
alexgoff, line 70: there's a context here for "pendingGroup", but I don't see a matching key in the translation strings, is it missing or is this context unused?

alexgoff (changes requested):
just one small note to check that an i18n string has a matching context, otherwise looks ok

public/localeStrings/en/translation.json
```diff
@@ -243,6 +243,16 @@
     "loading": "Activating your account…",
     "confirm_button": "Got it!"
   },
+  "facebook_sso": {
+    "success": "Success!",
+    "success_message": "Your account has been activated. Please log in to begin using the site as a registered user.",
+    "success_message_educator": "Your account has been activated, but it will need to be approved by an administrator before you can access educator content. You will receive an email when you have been approved.",
```
dananjohnson, line 249: @alexgoff here's the key that uses the `pendingGroup` context, falling back to just `success_message` as the default.

dananjohnson: See https://www.i18next.com/translation-function/context

alexgoff: oh, I didn't catch that `pendingGroup` was a variable and not a string, I was looking for a key like "success_message_pendingGroup" 🤯  everything looks good 😓

alexgoff (approved):

PR #39, opened 2023-08-10
[F] Wire up investigation progress bar

### LINKED ISSUE #38: Wire up progress bar

dananjohnson:
Resolves #38

## What this change does ##

This wires up the progress bar to data coming from the Craft API.

I talked to Kyle yesterday and we both felt like it would be better to represent investigation sections here and in the TOC by adding a bit of logic in the client rather than modeling it in the CMS. It seems like the main marker of a section is just that it ends with a save point. If that's not the case, or you find you need more functionality with sections, then we can look at defining it in Craft. But it seemed like not worth it due to the limited current needs.

## Notes for reviewers ##

You can use the `develop` branch of the API repo for this.

## Checklist ##

- [X] This change impacts several investigations (e.g. the change affects reuseed styles, widgets, pages, QAs, utility methods, etc.)


components/templates/InvestigationChildPage/index.tsx
```diff
@@ -16,8 +21,74 @@ const InvestigationChildPage: FunctionComponent<{
...
+      }
+    });
+
+    const mapped = sections
+      // filter out `undefined` from arrays
+      .filter((section) => section.some((el) => typeof el === "number"))
+      // create final section object
+      .map((section, index) => {
+        return {
+          name: `Section ${index + 1}`,
```
alexgoff, line 70: could this be i18nalized?

alexgoff (approved):
left one note about making a string internationalized, if you're pressed for time no worries, everything else looks fine. I think starting from using savepoints and growing as we need more functionality is fine and shouldn't be a big change to model in Craft if we need it.

PR #40, opened 2023-08-10
[F] Scaffold investigation content blocks

dananjohnson:
## What this change does ##

This PR scaffolds all of the content blocks in investigations so that we can begin building out the presentation for them. It defines GraphQL fragments and components for each block, and renders the typed data that's returned simply using `JSON.stringify()` to start. 

## Notes for reviewers ##

You can use the `develop` branch of the API repo for this. You'll need to create a investigation child page and some blocks to fully test.

## Checklist ##

- [X] This change is isolated to a specific page or Component (i.e. it's a one-off)

## Screenshots (optional) ##

Screenshot 2023-08-10 at 1 39 05 PM


alexgoff (approved):
added blocks to investigation pages, JSON output is as expected, looks good!

PR #63, opened 2023-10-19
EPO-8544 Color Filter Tool uses setup data from Color Filter Tool CMS Entry

blnkt:
JIRA: https://jira.lsstcorp.org/browse/EPO-8544

See https://github.com/lsst-epo/investigations-api/pull/36

## What this change does ##

Replaces the previous iteration of the Color Tool with a new version that takes all of its setup config from Craft.

## Notes for reviewers ##

This is in a very "just got it working and then stopped" state.  So, it does important thing, but probably does not yet do all of the things, nor does it maybe work 100% as expected in all contexts (2-col, in an Interaction Group, etc.).

## Testing ##

In the CMS, create a new Widget Entry with Type "Color Filter Tool".  Build it out with whatever settings and such as you like.  Go to an InvestigationChild page entry, add a Color Filter Tool Content Block, then select your new Color Filter Tool Widget entry from the Entry Filed in the Color Filter Tool Content Block, and save your work.  Go to the frontend and voila.  You should be able to interact with it as expected.  It should be styled as seen in Storybook.  You will see some errors in the console about not spreading key props.  They seem to be related to the widget as imported, and not work unique to `investigations-client`.


alexgoff (approved):
I was able to view the Color Tool that you created as an example and then create my own successfully so I think that portion is fine. 

I noticed 3 issues, all of which are outside of your branches scope:

1. the `CSS.supports` API used by `FilterImage` does not work in the server rendering pass. This component will need to be re-written for the export feature so this should be handled then [and I opened an issue for it](https://github.com/lsst-epo/epo-react-lib/issues/114)

2. the spread key prop, also a widget library issue
3. the GraphQL codegen really doesn't play nicely with TypeScript even after generating, looking at the code it seems impossible to extract useful types given how jumbled they are. Something to think about down the road I guess.

PR #65, opened 2023-10-24
[F] Table Content Block

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8565

## What this change does ##

Add Table Content Block to Craft and client

## Notes for reviewers ##

Edit or create a new table in Craft, check that the first row of the table is set as the header, check that enabling "isHeader" for any following row's first cell sets it as a header element as well, check that styling and content is as expected.

- 3 = "I barely need review on this"

## Checklist ##


components/content-blocks/InteractionGroupContainer/styles.ts
```diff
@@ -5,10 +5,15 @@ export const Heading = styled.h2`
   margin-block-end: ${fluidScale("2em", "1.5em")};
 `;
 export const InteractionGroup = styled.div`
+  --content-block-margin: ${fluidScale(
```
alexgoff, line 8: trying to standardize this into something reusable

components/content-blocks/Table/Table.tsx
```diff
@@ -0,0 +1,75 @@
...
+      return tableRow.map(({ cellContent = "", rowHeader = false }) => {
+        return {
+          children: cellContent,
+          ...(i > 0 && { isHeader: rowHeader }),
+        };
+      });
+    });
+  };
+
+  const rows = buildRows(Array.from(displayTable) as any);
```
alexgoff, line 52: it's a huge pain getting the types from GQL into something usable

components/content-blocks/Table/styles.ts
```diff
@@ -0,0 +1,19 @@
...
+  --table-negative: ${fluidScale(
+    "var(--PADDING_LARGE, 100px)",
+    "var(--PADDING_SMALL, 20px)"
+  )};
+
+  margin-inline: calc(var(--table-negative) * -1);
+
+  @media only screen and (min-width: ${token("BREAK_TABLET")}) {
+    --table-negative: 0;
+  }
```
alexgoff, line 18: for small screens, tables span to the edge of the screen

contexts/Progress.tsx
```diff
@@ -64,7 +64,8 @@ const getTwoColQuestionEntryIds = (block: object) => {
   const { columns } = block;
 
   const colBlocks = [...columns].map((col) => {
-    return [...col.children];
+    const { children = [] } = col;
```
alexgoff, line 67: encountered a bug here where `children` was undefined

blnkt (approved):

PR #67, opened 2023-10-26
[F] widget question type

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8592

## What this change does ##

Added a new "Widget" type of question in CraftCMS and the client that allows a widget to be paired with question texted and the widget's data saved and recalled.

## Notes for reviewers ##

Review using the API branch, client branch, and possibly new database dump.

## Testing ##

Create a new question with question text and associate a widget entry with it (any color tool widget is fine). Validate that it displays on the desired page, make a modification to the widget, check in local storage that the change is stored there, on reloading the page make sure that the data is properly recalled to the widget out of local storage.

## Checklist ##


components/content-blocks/Questions/styles.ts
```diff
@@ -16,7 +16,37 @@ export const QuestionList = styled.ol`
...
+      margin-block-start: var(--PADDING_SMALL, 20px);
+    }
+  }
+
   & > li + li {
     margin-block-start: var(--PADDING_SMALL, 20px);
   }
 `;
+
+export const QuestionLabel = styled.div`
```
alexgoff, line 30: since we are putting HTML in the question labels, including content like unordered lists that cannot be a child element of a `label` the label is now a simple text block. There is some styling here to massage everything together, particularly that the first child element must be displayed inline so it is next to the list number

components/content-blocks/TwoColumnContainer/TwoColumnContainer.tsx
```diff
@@ -26,7 +25,6 @@ const Fragment = graphql(`
         childblocks: children {
           __typename
           id
-          ...ColorFilterToolBlock
```
alexgoff, line 29: color filter tool is no longer a part of content blocks since it is the more complex configuration question-type widget

contexts/Progress.tsx
alexgoff, line 1: I encountered some issues with question numbers, I think relating to the nesting of questions, so I did a bit of work tightening up the Progress context while I figured it out.

contexts/Progress.tsx
```diff
@@ -73,49 +93,52 @@ const getTwoColQuestionEntryIds = (block: object) => {
...
+      case "contentBlocks_questionBlock_BlockType": {
+        return getQuestionEntryIds(block);
+      }
+      case "contentBlocks_twoColumnContainer_BlockType": {
+        return getTwoColQuestionEntryIds(block);
+      }
+      case "contentBlocks_group_BlockType": {
+        const { group } = block;
+
+        return getPageQuestionEntryIds(group).flat();
```
alexgoff, line 110: instead of additional work here, just do it recursively and flatten the result

blnkt (approved):

PR #71, opened 2023-10-30
[F] Inline questions

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8593

## What this change does ##

Adds client wire up for inline questions, includes new typings for stored answers and some changes to how answers are sent to the server (as JSON)

## Notes for reviewers ##

Use the new database uploaded to dumps on 10/30 and check the inline questions (pg. 10 is a good starting point), and verify them as well in Craft that the entries can be created, saved, and added to pages.

## Testing ##

- Verify inline questions have correct client display on pg. 10.
- Complete questions, leave the page or investigation, and return. Verify that the questions load with stored answers correctly.
- Progress to a save point and verify that the saved answers go to the server. 

## Checklist ##


components/educator-schema/saveAnswersAction.ts
```diff
@@ -35,7 +35,12 @@ export default async function saveAnswers(
     return "statusError";
   }
 
-  const answerSet = Object.values(answers);
+  const answerSet = Object.values(answers).map(({ data, ...values }) => {
```
alexgoff, line 38: @kylepratuch I wanted your input here on stringifying the `data` property before putting it into the query. We talked about the DB storing all values as a string type, since the typing of `data` is very loose for the different question types (it can be a string, array of strings, generic object, mildly typed object), I opted to stringify all the data rather than trying to adapt the limited GQL types to fit.

components/questions/InlineQuestion/index.tsx
```diff
@@ -72,9 +69,18 @@ const InlineQuestion: FunctionComponent<InlineQuestionProps> = ({
...
           <Input
-            key={id}
-            onChangeCallback={callback}
-            {...{ ...props, isDisabled, id }}
+            key={partId}
+            id={partId}
+            onChangeCallback={(value: string | string[]) =>
+              onChangeCallback &&
+              onChangeCallback(
+                { ...(data as InlineQuestionData), [partId]: value },
```
alexgoff, line 77: updates the `data` object with a value using the inline question part's ID as the key.

components/questions/SimpleQuestion/styles.ts
```diff
@@ -0,0 +1,29 @@
+"use client";
+import styled from "styled-components";
+
+export const SimpleContainer = styled.li`
+  & > * + * {
+    margin-block-start: var(--PADDING_SMALL, 20px);
+  }
+`;
+
+export const QuestionLabel = styled.div`
```
alexgoff, line 10: co-locate these styles with the `SimpleQuestion` component so they don't pollute the inline questions

blnkt (approved):

PR #73, opened 2023-11-02
[F] Reference Modal Block

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8594

## What this change does ##

Adds support for the Reference Modal entry content block. Includes a significant refactor of modals in the application along with support for a standalone reference entry page. 

## Notes for reviewers ##

Check this branch and the corresponding API branch. Use the `db_ctu_20231102.sql` DB on GC

- 6

## Testing ##

Test the following
- Reference entries exist as a new section and entry type
- Reference entries accept content blocks as children
- Adding a reference entry to a page shows a button
- Clicking the button opens the reference entry content in a modal (optional: a loader shows while the modal is open but content is not fetched yet)
- When the modal is open the URL is re-written to /[investigation]/reference/[entry slug]
    - should the "reference" be a child of the [investigation] param? They aren't tied to any investigation so they could go at the root as /reference/[entry slug] but I'm not sure if there's any auth implications in there
- When the modal is open, either cmd+left arrow navigates back to the previous URL or "escape" closes the modal. 
- If the page is refreshed with the modal open, or the URL is copied and pasted, the content shows in a new page.

## Checklist ##


app/[locale]/[investigation]/@modal/default.tsx
```diff
@@ -0,0 +1,3 @@
+export default function Default() {
```
Comment (alexgoff) [inline app/[locale]/[investigation]/@modal/default.tsx:1]: these `default` pages are needed for parallel routes where routing does not exist at the top level

app/[locale]/[investigation]/layout.tsx
```diff
@@ -43,6 +44,7 @@ const InvestigationLandingLayout: (
   props: PropsWithChildren<InvestigationLandingProps>
 ) => Promise<JSX.Element> = async ({
   children,
+  modal,
```
Comment (alexgoff) [inline app/[locale]/[investigation]/layout.tsx:47]: for parallel routes, a named child component is passed as a prop and then rendered alongside `children` whereever we choose it.

app/[locale]/[investigation]/reference/[slug]/page.tsx
```diff
@@ -0,0 +1,49 @@
...
+    },
+  });
+
+  const { entry } = data || {};
+
+  if (!entry || entry.__typename !== "referenceModals_default_Entry") {
+    notFound();
+  }
+
+  return <ReferenceContentPage data={entry} site={site} />;
```
Comment (alexgoff) [inline app/[locale]/[investigation]/reference/[slug]/page.tsx:46]: this is the standalone reference page routing

app/[locale]/layout.tsx
```diff
@@ -95,7 +95,10 @@ const RootLayout: (
   return (
     <html lang={locale}>
       <head></head>
-      <body className={SourceSansPro.variable}>
+      <body
+        className={SourceSansPro.variable}
+        style={{ "--scrollbar-width": "calc(100vw - 100%)" }}
```
Comment (alexgoff) [inline app/[locale]/layout.tsx:100]: added this so that non-fullscreen modals can get the width of the screen minus any visible scrollbars

components/content-blocks/Modal/index.tsx
alexgoff, line 1: rewrote this entire block, the previous setup didn't reflect how this is intended to work.

components/factories/ContentBlockFactory/ContentBlockFactory.tsx
```diff
@@ -28,26 +29,41 @@ const Fragment = graphql(`
...
 
   if (!Block) return null;
-  return <Block data={data} site={props.site} pageId={props.pageId} />;
+  return (
+    <Block
+      data={data}
+      site={site}
+      pageId={pageId}
+      isOpen={isOpen}
+      hasModal={hasModal}
```
alexgoff, line 64: added these so we can control if content blocks render in their modal-open state, and if they have modal controls (for reference modals we want them "open" but NOT showing their own modal controls or we get duplicate modals)

components/hoc/withModal/withModal.tsx
```diff
@@ -2,31 +2,40 @@ import { ComponentType, useState, useRef, FunctionComponent } from "react";
...
   const WithModal = (props: T) => {
-    const { title } = props;
+    const { title, hasModal = true } = props;
+
     const [isOpen, setIsOpen] = useState(false);
     const modalRef = useRef(null);
+    const uid = useUID();
+
+    // escape for content that is already within a modal (reference pages)
+    if (!hasModal) {
```
alexgoff, line 23: to escape the `withModal` wrapper on images and widgets, we set `hasModal` to false when rendered inside of a reference modal

components/hoc/withModal/withModal.tsx
```diff
@@ -45,55 +54,35 @@ function withModal<T extends BaseContentBlockProps>(
...
-                isOpen,
-                openModal,
-                closeModal,
-                labelledById: title && isOpen ? titleId : undefined,
-              }}
-            />
-          </Styled.Modal.ComponentWrapper>
-        </div>
-      </Styled.Modal.Dialog>
+      <Modal.Base
```
alexgoff, line 61: the rest is just following refactor to move modal functionality into a reusable base

components/layout/Modal/Base/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    return (
+      <Styled.Dialog open={isOpen} ref={ref}>
+        <Styled.Backdrop open={isOpen} />
+        <div
+          role={isOpen ? "dialog" : "generic"}
+          aria-modal={isOpen}
+          aria-labelledby={labelledById}
+          id={id}
+        >
```
alexgoff, line 32: this modal base handles everything up to the dialog container with focus trap and keyboard events

components/templates/ReferenceContentPage/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+const ReferenceContentPage: FunctionComponent<{
+  data: FragmentType<typeof Fragment>;
+  site: string;
+  isInModal?: boolean;
+}> = ({ site, isInModal = false, ...props }) => {
+  const data = useFragment(Fragment, props.data);
+  const { title, id } = data;
+
+  return (
+    <Styled.ContentBlocks paddingSize="none" width="wide" bgColor="transparent">
```
alexgoff, line 27: this is the content that goes either into the reference page or reference modal

blnkt (changes requested):
Update as needed based on changes to analogous [investigations-api refactor](https://github.com/lsst-epo/investigations-api/pull/42)

PR #76, opened 2023-11-08
[C] UI fixes for pager + loaders

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8627

## What this change does ##

Fixes the pager flowing outside of the screen and weird horizontal scroll. Generally cleans up some of the lower level styles and HTML structure. Finally adds some additional loaders for better page transitions.

## Notes for reviewers ##

- 2

## Testing ##

Go through the Coloring the Universe investigation on several screen sizes using the browser emulator. Pager should have space above it now, separating the content visually. Pager should be centered and remain at the size of the root container. Weird horizontal overflow should be gone now.

## Checklist ##


app/[locale]/[investigation]/[...uriSegments]/layout.tsx
```diff
@@ -62,7 +63,7 @@ const UriSegmentsLayout: (
   return (
     <ProgressProvider pages={childPages} currentPageId={id}>
       <Header />
-      {children}
+      <Styled.Main>{children}</Styled.Main>
```
Comment (alexgoff) [inline app/[locale]/[investigation]/[...uriSegments]/layout.tsx:66]: the `main` element should group the content of the page and exclude re-used nav and UI elements

app/[locale]/layout.tsx
```diff
@@ -95,21 +95,18 @@ const RootLayout: (
   return (
     <html lang={locale}>
       <head></head>
-      <body
-        className={SourceSansPro.variable}
-        style={{ "--scrollbar-width": "calc(100vw - 100%)" }}
-      >
+      <Body className={SourceSansPro.variable}>
```
Comment (alexgoff) [inline app/[locale]/layout.tsx:98]: the styled `body` is moved up to the highest level so we don't have redundant `body` and `Body` elements

components/page/Body/index.tsx
```diff
@@ -0,0 +1,17 @@
+import * as Styled from "./styles";
+import { FunctionComponent, PropsWithChildren } from "react";
+
+const Body: FunctionComponent<PropsWithChildren<{ className?: string }>> = ({
+  className,
+  children,
+}) => {
+  return (
+    <Styled.Body className={className}>
```
alexgoff, line 9: `Body` now contains the HTML body and a `div` container to limit maximum size.

blnkt (approved):

PR #78, opened 2023-11-14
[EPO-8616] Homepage ContentBlocks (including InvestigationGrid)

blnkt:
JIRA: https://jira.lsstcorp.org/browse/EPO-8616>

## What this change does ##

Homepage single has its own special set on ContentBlocks which includes:
 - TextBlock
 - ImageBlock
 - new InvestigationGrid

## Notes for reviewers ##

This PR uses the `investigations-api` PR: https://github.com/lsst-epo/investigations-api/pull/47.  

## Testing ##

Go into the CMS and update the homepage with some content and stuff.  Check that image, text, and investigationGrid blocks all render as expected on the frontend.


components/content-blocks/InvestigationGrid/InvestigationGrid.tsx
```diff
@@ -0,0 +1,89 @@
...
+      <Grid columns={3} tablet={3}>
+        {investigations.map(
+          ({ id, title, rawImage, url, isInactive = false, status = "active" }) => {
+            const image =
+              rawImage?.length > 0 ? imageShaper(site, rawImage[0]) : null;
+            // return <Tile key={id} title={title} image={image} link={link} />;
+            return (
+              <Styled.MixedLink
+                key={id}
+                aria-labelledby={id}
```
alexgoff, line 62: this `labelledby` can be removed if `role="presentation"` is passed to the Image, then the text will be read as the link element's content. I don't think the images will ever contain any visual information that needs to be conveyed by alt.

alexgoff (changes requested):
few little pieces of cleanup, otherwise everything is acting and working as it should! The API branch is good to go.

PR #79, opened 2023-11-14
[F] Wire up review page

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8669

## What this change does ##

Wired up review page to display the user's saved answers.
- Separated pages, questions, answers, and progress contexts
- Moved page navigation queries up to [investigation] level so they are queried once then held in a context
- Created a display-only Color Filter Tool container that can be used in questions and reviews with minimal config
- Refactored a bunch of the review page display components to remove crufty components in the middle (a later task would be applying the same to the questions themselves)
- This does not include work on print styling or XLSX output of review data.

## Notes for reviewers ##

Review change on the frontend, there is no API branch or DB update for this PR.

- 7

## Testing ##

- Complete one or several investigation pages of questions
- On the final page of the investigation, a button link should display that navigates to the Review page
- On the review page all questions should be displayed, with answers for questions that were answered.
- Placeholder text should be rendered where answers were not provided
- The previous page should be navigable from the review page

## Checklist ##

## Screenshots (optional) ##

image
image


components/factories/ReviewFactory/index.tsx
```diff
@@ -1,31 +1,36 @@
...
+  type: string;
+  number: number;
+  value?: AnswerData;
   config: any;
 }
 
 const QUESTION_MAP: Record<string, ComponentType<any>> = {
-  simple: SimpleReview,
-  inline: InlineReview,
+  text: Review.Text,
```
alexgoff, line 14: no more Simple vs. Inline factory in the middle, just go straight to the components since that reflects the data model in the CMS

components/page/Pager/Pager.tsx
```diff
@@ -29,39 +74,48 @@ const Pager: FunctionComponent<PagerProps> = ({
...
-        href={isPreviousDisabled ? "#" : leftLink}
+        href={isPreviousDisabled ? "#" : previousPage}
         aria-disabled={isPreviousDisabled}
       >
         {leftText || t("pager.previous")}
       </Styled.PagerButton>
       <Styled.PageCount>
         {t("pager.page-count", {
-          current: currentPageNumber,
+          current: Math.min(currentPageNumber, totalPages),
```
alexgoff, line 113: make sure we never go beyond the total number of pages, even on the review page

components/templates/InvestigationChildPage/index.tsx
```diff
@@ -15,112 +15,6 @@ const Fragment = graphql(`
       ...ContentBlockFactory
     }
     hasSavePoint
-    prev(section: "investigations") {
```
alexgoff, line 18: remove all of this prev/next/parent stuff from individual pages now that we have an overarching context

contexts/Pages.tsx
alexgoff, line 1: most of this is copied out of Progress

blnkt (approved):
Had to clear my localStorage and hard reload to get the review page to load correctly, but once I did everything worked as advertised.  Great wrok.

PR #80, opened 2023-11-14
[EPO-8680] nextjs upgrade + assoc deps & config

blnkt:
JIRA: https://jira.lsstcorp.org/browse/EPO-8680

## What this change does ##

Upgrades nextjs to 14 (and types to suit).  Also updates to node 20.2.0 because next 14 requires at least 18, and 20 is the current LTS release.  Also a couple tweaks to imports and config as a result of the upgrade. 

## Notes for reviewers ##

`.node-version` and dockerfile have both been updated to reflect the new node version used in this repo, however you will need to install this version for local dev.  If you have `nodenv` already, then you can accomplish this by running `nodenv install` from the root of this project.

## Testing ##

Click through the frontend and check js console and terminal logs for anything weird (via `yarn dev`).  Also try to build a build and make sure it does so free of errors (`yarn bs` will build and then serve at localhost:3000)


alexgoff (approved):
no issues getting new dependencies install, client is up and running, looks good to me

PR #81, opened 2023-11-16
[F Save work flow

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8656

## What this change does ##

A few sentences describing the outcome of the change (“fixes a thing”; "adds the stuff") and how you tackled it (“added a prop to prevent lazyloading in the presence of barbara walters”).  How successful is the change?  And how are you measuring that success?

## Notes for reviewers ##

Client branch only, no updates to API or database. 

- 3

## Testing ##

- Start either logged in or out and complete up to Colorizing the Images in Coloring the Universe
- Click the save button
- A toast should appear that indicates save status. If successful but not logged in, additional buttons should appear that direct users to log in
- Error statuses can be tested by editing the save actions to always return errors.

## Checklist ##

## Screenshots (optional) ##

image


components/answers/SaveForm/SaveForm.tsx
```diff
@@ -1,97 +1,131 @@
...
-            `${investigationId}_answers`
-          );
-          if (
-            !storedAnswers ||
-            !Object.values(JSON.parse(storedAnswers)).length
-          )
-            return setStatus("emptyError");
+  if (!investigationId) return null;
+
+  const saveAnswers = async () => {
```
alexgoff, line 38: this `saveAnswers` method needs to be extracted to a server-only action. It works as-is but leaves an exception that gets logged into the DOM because form actions are intended to be serverside. Refactoring this presents some issues since local storage is not available so the clientside work must be done and then bound to the action. Leaving as-is for now but would like to fix.

components/answers/SaveForm/SaveForm.tsx
```diff
@@ -1,97 +1,131 @@
...
+        <Styled.Checkpoint>
+          <Styled.CheckpointIcon>
+            <IconComposer icon="Thumbtack" size="1.5em" />
+          </Styled.CheckpointIcon>
+          <Styled.CheckpointDivider />
+          {t("answers.save_form.checkpoint")}
+        </Styled.Checkpoint>
+        <SaveButton />
+      </Styled.Form>
+      <Toaster
```
alexgoff, line 119: toaster lives outside the form so buttons inside of it do not trigger the form action

components/layout/Toaster/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <Styled.ToastContent>
+          {icon && (
+            <Styled.ToastIcon>
+              <IconComposer size="1em" icon={icon} />
+            </Styled.ToastIcon>
+          )}
+          <Styled.ToastText>
+            <h3>{title}</h3>
+            <p>{text}</p>
+            {additionalContent}
```
alexgoff, line 90: overall I'm moderately happy with this, toasts are difficult because they are not anywhere close to an agreed upon web standard. This toast will announce the new message using the `status` role but buttons within it can't be keyboard navigated because focus does not move to the toast. Moving focus to the toast is not desirable since after saving users will want to continue to the navigation.

PR #83, opened 2023-11-28
[EPO-8793] Narrower max container width 

blnkt:
JIRA: https://jira.lsstcorp.org/browse/EPO-8793>

## What this change does ##

Wherever `Container` is used gets the `width="narrow"` prop to approximate the max container width shown in design.  It's still a bit wider than in the design (by around 100px) but feels....fine?

## Notes for reviewers ##

Chances the `Container` on homepage, investigation landing page, investigation child page, review page, and reference content page.

## Testing ##

Please visually inspect on all effected page layouts at desktop and mobile screen widths.


alexgoff (approved):
it looks good here, checked across a range of screen sizes and I agree that the narrower width appears better especially with larger graphics.

PR #84, opened 2023-11-29
[C] menu content

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8670

## What this change does ##

Fleshes out menu content with acknowledgements, sharing, and help. See corresponding API branch.

## Notes for reviewers ##

Review this change on both the client and API. 

Run `yarn` to pick up latest version of libs that makes updates to slideout menu and sharing buttons

In Craft a new "Help URL" should be available in the globals, and under Investigation parent entries an acknowledgements field should be available.

-5 

## Testing ##

- Add test acknowledgements to Coloring the Universe in Craft
- Add a test help URL to the globals in Craft
- Open the menu in an investigation and verify acknowledgements
- Open the menu in an investigation and verify help link
- Open the menu in an investigation and verify sharing

## Checklist ##


blnkt (changes requested):
The helpUrl button needs to be a link.  Right now it's a `MenuItem` button which gets an `href` and a `type="link"` but I don't believe that is a thing.  I believe it should be an `<a>` styled to look like a button.  I don't know if it makes more sense to implement in `MenuItem` given some prop, or if in the investigation's Menu as like a one-off that's shaped like a `MenuItem`.

components/page/Menu/Menu.tsx
```diff
@@ -25,6 +35,26 @@ const Menu: FunctionComponent<MenuProps> = ({ isOpen, onCloseCallback }) => {
...
+            onCloseCallback={() => setIsSubMenuOpen(false)}
+          />
+        )}
+        {helpUrl && (
+          <MenuItem
+            type="link"
+            href={helpUrl}
+            target="__blank"
+            text="Help"
+            icon="QuestionCircle"
```
blnkt, line 55: Is not quite doing what we want it to

blnkt:
nm.  Reviewer was in error.  PR is A-ok.

PR #88, opened 2023-12-01
[C] Fix login/activation flows

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8800

## What this change does ##

Fix login and activation flow issues, changes include:

- activation link now works
- user status is updated from pending to active when activation is completed
- change start button to go to first investigation page when authed
- show currently authed user
- show additional message if user is authed but pending
- add create account button when not authed

## Notes for reviewers ##

client side only, no changes to API. 

## Testing ##

- start logged out
- create a new account
- see that after creation, account is in a pending state without logged in investigation status (user will have to log out or activate to start investigation)
- click activation link and see that activation modal appears, and user is moved to active status
- see that investigation start is now available

## Checklist ##

## Screenshots (optional) ##
unauthed
image

authed but pending
image

authed
image


app/[locale]/layout.tsx
```diff
@@ -101,7 +102,9 @@ const RootLayout: (
             <GlobalStyles includeFonts={false} />
             <I18NextClientProvider locale={locale}>
               <GlobalDataProvider data={globalData}>
-                {children}
+                <AuthDialogManagerProvider>
```
Comment (alexgoff) [inline app/[locale]/layout.tsx:105]: move the auth dialog provider up higher so it can be accessed in different areas of the app

components/answers/SaveForm/SaveForm.tsx
```diff
@@ -96,8 +96,8 @@ const SaveForm: FunctionComponent<{
           text: toastText[status],
           additionalContent: !user ? (
             <>
-              <Button isBlock>Log in</Button>
-              <Button isBlock>Create an Account</Button>
+              {/* <SignIn />
```
alexgoff, line 99: these work but they're kind of weird and we need to figure out how to handle things like users creating an account, being put in pending status, and left on the investigation page. Pending users can't save to the database so I have been disabling access while pending.

which does then raise the question in my mind, do we need to be redirecting pending users? it's just disabled in the UI, users can still route to it.

components/auth/dialogs/Activate/actions.ts
```diff
@@ -25,9 +22,13 @@ export async function activate(
   });
 
   if (data?.activateUser) {
-    if (pathToRevalidate) {
-      revalidatePath(pathToRevalidate);
+    const token = cookies().get("craftRefreshToken")?.value; // check if there is a user already logged in
```
alexgoff, line 25: if there is an account logged in, refresh the cookies after activation. If it is the user's account then they'll now be able to progress, if it's somebody else's account nothing will happen.

components/templates/HomePage/index.tsx
```diff
@@ -60,7 +60,7 @@ export default function HomePage(props: {
         (block, i) =>
           block && <ContentBlockFactory key={i} site={data.site} data={block} />
       )}
-      {/* {props.children} */}
+      {props.children}
```
alexgoff, line 63: need this for auth dialogs

PR #91, opened 2023-12-04
[C] add French support

alexgoff:
For JIRA: "EPO-8671 #IN-REVIEW #comment add French support"

JIRA: https://jira.lsstcorp.org/browse/EPO-8671

## What this change does ##

Added French support to client, extended `getSite` helper to determine the correct site in a beyond-binary situation. Added French translations file and cleaned up existing translations files.

## Notes for reviewers ##

Grab this branch and the API. Check both in the API (that Craft has a new French site available) and check in the client that French is working

- 2
- 
## Testing ##

Go to /fr/coloring-the-universe and observe that it can be navigated to, enter the investigation and observe that inside of the slideout menu "French" is now selected as an option. Create some new French content in the CMS and


.storybook/i18next.js
```diff
@@ -12,7 +13,7 @@ i18n
   .init({
     debug: true,
     lng: "en",
-    fallbackLng: "en",
+    fallbackLng,
```
alexgoff, line 16: keep Storybook in sync with the rest of the app

app/[locale]/[investigation]/layout.tsx
```diff
@@ -21,14 +22,30 @@ export interface InvestigationLandingProps {
...
+  query InvestigationMetadata($site: [String], $uri: [String]) {
+    entry(site: $site, uri: $uri) {
+      ... on investigations_investigationParent_Entry {
+        title
+      }
+    }
+  }
+`);
 
 export async function generateMetadata({
```
Comment (alexgoff) [inline app/[locale]/[investigation]/layout.tsx:35]: long overdue that the metadata is updated here

components/templates/ErrorPage/index.js
```diff
@@ -2,19 +2,16 @@
...
-    <Body title={title}>
-      <Container>
-        <h1>{title}</h1>
-        <div
-          className="c-content-rte"
-          dangerouslySetInnerHTML={{ __html: text }}
-        />
-      </Container>
-    </Body>
+    <Container>
```
alexgoff, line 8: was getting a bunch of errors every time I hit a page not found and remove the body tag from this page

helpers/index.ts
```diff
@@ -14,7 +14,7 @@ import type { GlobalsQueryQuery } from "gql/public-schema/graphql";
 type Categories = GlobalsQueryQuery["categories"];
 
 export const getSite = (locale: string = fallbackLng) =>
-  locale === fallbackLng ? "default" : "es";
+  locale === fallbackLng ? "default" : locale;
```
alexgoff, line 17: if it's not the fallback language, then the site we want is the locale key

public/localeStrings/en/translation.json
alexgoff, line 1: I removed a bunch of rubin-obs-client translations that weren't serving a purpose in here, I think it is better to pare down to what we need so our translators aren't needlessly translating content.

public/localeStrings/es/translation.json
alexgoff, line 1: there was no Spanish content in here so I removed it so that it doesn't erroneously show up as populated. Now we can enable debug mode and track what is/isn't translated.

public/localeStrings/fr/translation.json
alexgoff, line 1: high school french verified by translate 👍🏻

blnkt (approved):

PR #101, opened 2023-12-14
[EPO-8802] Google & Facebook auth fixes

blnkt:
JIRA: https://jira.lsstcorp.org/browse/<ID>

## What this change does ##

A few sentences describing the outcome of the change (“fixes a thing”; "adds the stuff") and how you tackled it (“added a prop to prevent lazyloading in the presence of barbara walters”).  How successful is the change?  And how are you measuring that success?

## Notes for reviewers ##

Where should the reviewer look for the change?  Do they review the change on the frontend?  How extensive is the change?  Does it affect Components used in a variety of contexts?  Anything you'd like reviewers to pay particular attention to or questions you have about the change.

- 1 = "I barely need review on this"
- 10 = "I think it works, I can explain how, but only over screenshare"

## Testing ##

How did you test this change?  Can the reviewer test this change for themselves?

If you wrote new tests in the change it's fine to just put "New tests in PR". If you haven't added any tests but you think your change is covered by existing tests you can say "Existing tests". And if you're not sure how to test it, say that.

## Checklist ##


components/auth/AuthDialogs/AuthDialogs.tsx
```diff
@@ -13,7 +13,7 @@ export default function AuthDialogs({
       <Dialog.SelectProvider />
       <Dialog.SetPassword />
       <Dialog.SignIn />
-      <Dialog.SignInFacebook />
+      {/* <Dialog.SignInFacebook /> */}
```
blnkt, line 16: Disables facebook stuff for now.

components/auth/buttons/FacebookSSO/FacebookSSO.tsx
```diff
@@ -1,9 +1,8 @@
 "use client";
 
 import { useState } from "react";
-import { Button } from "@rubin-epo/epo-react-lib";
 import { useTranslation } from "react-i18next";
-import { getOauthUrl } from "./actions";
+import FacebookLogin from '@greatsumini/react-facebook-login';
```
blnkt, line 5: Even though is disabled, in my messing around with stuff it feels like a package like this is going to simplify our lives

components/auth/buttons/GoogleSSO/Button.tsx
```diff
@@ -46,10 +51,23 @@ export default function SSOButton({
...
+  //   </Button>
+  // );
   return (
-    <Button onClick={() => login()} styleAs="tertiary">
-      {t("sign_in.continue_with_google")}
-    </Button>
+    <GoogleLogin
+      onSuccess={handleSuccess}
+      onError={handleError}
+      locale={language}
```
blnkt, line 63: We can't much customize the text in the button but we can translate it which is nice

components/auth/buttons/GoogleSSO/Button.tsx
```diff
@@ -46,10 +51,23 @@ export default function SSOButton({
...
+      shape="rectangle"
+      size="large"
+      text="continue_with"
+      theme="outline"
+      type="standard"
+      width={buttonWidth}
```
blnkt, line 69: Can't style directly but made do with the available props for styling

components/auth/buttons/GoogleSSO/GoogleSSO.tsx
```diff
@@ -12,11 +14,25 @@ export default function GoogleSSO({
   className: string;
   onError: () => void;
 }) {
+  const [buttonWidth, setButtonWidth] = useState();
+
+  const { ref } = useResizeObserver({
+    onResize: ({ width }) => {
+      setButtonWidth(width);
+    },
+  });
```
blnkt, line 23: the google button needs an explicit width (because the year is 2010 or something) so using our resizeObserver to like fake responsiveness.

components/auth/dialogs/SelectProvider/styles.ts
```diff
@@ -4,22 +4,28 @@ import AuthButtons from "@/components/auth/buttons";
...
+  justify-content: center;
+  width: 100%;
   margin-top: 50px;
 `;
 
+export const WidthConstrainer = styled.div`
+  max-width: 400px;
+  width: 100%;
+`;
+
```
blnkt, line 16: because the google button has a hardcoded max-width of 400px I decided to just resize the whole inside of the auth modals that its featured in so that all of the UI kind matches width and feels more intentional than it looked when just the google button was different.

components/auth/dialogs/SignIn/styles.ts
```diff
@@ -3,9 +3,17 @@ import Submit from "@/components/form/Submit";
...
+  justify-content: center;
+  width: 100%;
   padding-top: 13px;
 `;
 
+export const WidthConstrainer = styled.div`
+  max-width: 400px;
+  width: 100%;
+`;
+
```
blnkt, line 16: because the google button has a hardcoded max-width of 400px I decided to just resize the whole inside of the auth modals that its featured in so that all of the UI kind matches width and feels more intentional than it looked when just the google button was different.

PR #107, opened 2023-12-20
[C] add error parser to signup

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8841

## What this change does ##

"error" status was not reflecting properly in signup. Address this bug and additionally put an error parser in that interprets the GraphQL errors into a translateable message.

## Notes for reviewers ##

Client-only

- 3

## Testing ##

Test the following cases:

- Create a new account with an email that is already signed up
- Create a new account with an invalid email (ex. a@b.)
- Create a new account with a password that does not meet the minimum length (frontend validation was added on this as well so the error catch may be unnecessary)

## Checklist ##

### DIFF
--- a/components/auth/dialogs/SignUp/SignUp.tsx
+++ b/components/auth/dialogs/SignUp/SignUp.tsx
@@ -1,7 +1,10 @@
 "use client";
 
 import { useState } from "react";
-import { BasicModal, Input } from "@rubin-epo/epo-react-lib";
+import BasicModal from "@rubin-epo/epo-react-lib/BasicModal";
+import FormButtons from "@rubin-epo/epo-react-lib/FormButtons";
+import Button from "@rubin-epo/epo-react-lib/Button";
+import Submit from "@/components/form/Submit";
 import { useAuthDialogManager } from "@/contexts/AuthDialogManager";
 import { registerEducator, registerStudent } from "./actions";
 import { useTranslation } from "react-i18next";
@@ -11,17 +14,14 @@ export default function SignUp() {
   const { active, pendingGroup, openModal, closeModal } =
     useAuthDialogManager();
 
-  const [status, setStatus] = useState<"error" | null>(null);
+  const [error, setError] = useState<string | null>(null);
 
   const { t } = useTranslation();
 
   function getTitle() {
-    switch (status) {
-      case "error":
-        return t("register.error");
-      default:
-        return t("register.header", { context: pendingGroup });
-    }
+    if (error !== null) return t("register.error");
+
+    return t("register.header", { context: pendingGroup });
   }
 
   return (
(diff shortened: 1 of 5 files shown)


components/auth/dialogs/SignUp/SignUp.tsx
```diff
@@ -11,17 +14,14 @@ export default function SignUp() {
   const { active, pendingGroup, openModal, closeModal } =
     useAuthDialogManager();
 
-  const [status, setStatus] = useState<"error" | null>(null);
+  const [error, setError] = useState<string | null>(null);
```
alexgoff, line 17: store the error message instead of just the status

components/auth/dialogs/SignUp/SignUp.tsx
```diff
@@ -46,83 +46,75 @@ export default function SignUp() {
                 openModal("statusPending");
               }
             }
+
+            setError(null);
```
alexgoff, line 50: resets the error on success

components/auth/dialogs/SignUp/SignUp.tsx
```diff
@@ -46,83 +46,75 @@ export default function SignUp() {
...
-
-          setStatus(null);
         }}
       >
-        <Styled.InputWrapper>
-          <Styled.Label htmlFor="signUpEmail">
-            {t("form.email_required")}
-          </Styled.Label>
-          <Input
+        <Styled.Label>
```
alexgoff, line 56: did some stylistic refactor, inputs are now wrapped in labels instead of being associated by ID

components/auth/dialogs/SignUp/SignUp.tsx
```diff
@@ -46,83 +46,75 @@ export default function SignUp() {
...
+          {t("form.password_required")}
+          <Styled.Instructions>
+            {t("form.create_password_instructions")}
+          </Styled.Instructions>
+          <Styled.Input
             name="password"
             id="signUpPassword"
             type="password"
             autoComplete="current-password"
+            minLength={6}
```
alexgoff, line 94: frontend validation of minimum length

components/auth/dialogs/SignUp/actions.ts
```diff
@@ -1,9 +1,40 @@
...
+import { cookies } from "next/headers";
 import { graphql, useFragment } from "@/gql/public-schema";
 import { AuthFragmentFragmentDoc } from "gql/public-schema/graphql";
+import { CombinedError } from "@urql/core";
+import { mutateAPI } from "@/lib/fetch";
+import { serverTranslation } from "@/lib/i18n";
+import { setAuthCookies } from "@/components/auth/serverHelpers";
+import { fallbackLng } from "@/lib/i18n/settings";
+
+async function getErrorMessage(error: CombinedError): Promise<string> {
```
alexgoff, line 11: do some interpretation of the error message we get from GraphQL and make a new error message

blnkt (approved):
super.  See jira, I think should take this approach to other auth dialogs with errors sometime in the future: https://jira.lsstcorp.org/browse/EPO-8901

Addiitonally, this SignUp set of custom error messaging should handle `[graphql] email "student@lsst.org" has already been taken`.

PR #108, opened 2023-12-20
[C] use display name instead of title

alexgoff:
For JIRA: "EPO-8859 #IN-REVIEW #comment use display name instead of title"

JIRA: https://jira.lsstcorp.org/browse/EPO-8859

## What this change does ##

Uses the new "Display Name" field for widget titles instead of "Title"

## Notes for reviewers ##

Client and API changes

- 2

## Testing ##

- First see that with no changes in the CMS, the titles still appear as they did before ("Color Filter Tool" or "Rubin Color Filter Tool")
- Add a display name to any widget and see that it displays in the client instead of the entry title

## Checklist ##


blnkt (approved):
yep yep

PR #114, opened 2023-12-26
[F] Equation Block

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8891

## What this change does ##

Add an Equation block that renders LaTeX to MathML

## Notes for reviewers ##

Client + API. Here are some equations to try adding to the blocks:

```latex
\begin{equation}\begin{split}DM&=(m-M)\\&=(m-(-19.4))\\&=(m+19.4)\end{split}\end{equation}
```

```latex
d=10^{\frac{\left(DM+5\right)}{5}}
```

```latex
1\space\text{parsec}=3.26\space\text{light-years}
```

- 3

## Testing ##

Add equation content blocks in Craft and populate with valid LaTeX, check that they are rendered correctly frontend. Additionally can check that invalid LaTeX does not render by supplying incorrect markup and observing the error message rendered instead.

## Checklist ##


blnkt (approved):
👍 🔢

PR #116, opened 2024-01-04
[F] Formalize toaster setup

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8937

## What this change does ##

Formalizes the toaster that was created for the save answers form using `react-hot-toast` to create a reusable API for future toasts

## Notes for reviewers ##

Client-only change.

- 2

## Testing ##

- Visit a page with a save form, click "save", and observe either a success or error toast appear then disappear after a short duration.
- Repeat the same but click to close
- Try adding your own toast to a page: 
```tsx
import { toast } from 'react-hot-toast'
import Toaster from '@/components/layout/Toaster'

function Component() {

    return <><button onClick={() => toast("I'm a toast")}/><Toaster /></>
}
```

## Checklist ##


components/layout/Toaster/index.tsx
```diff
@@ -1,103 +1,65 @@
...
-
-      setTimeout(() => onCloseCallback && onCloseCallback(), toastDelay);
-    }
-  };
-
-  const { title, icon, text, additionalContent } = message || {};
-
+const Toaster = () => {
   return (
     <Styled.Toaster
```
alexgoff, line 8: set up a bunch of default styling and behaviors for our toaster here

app/[locale]/layout.tsx
Comment (alexgoff) [inline app/[locale]/layout.tsx:1]: originally the toaster was going to live here and I rearranged some imports in the process, nothing ended up changing

blnkt (approved):
👍 🍞

PR #119, opened 2024-01-05
[F] Fix server rendering + color option values

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8855

## What this change does ##

Adds the release candidate version of the `epo-widget-lib` with fixes for server rendering differences and fixes a bug where color option labels were being passed to values, leading to incorrect renders on client + server.

## Notes for reviewers ##

Review on client

- 4

## Testing ##

Use a test image like the RGB image filter and set all values to 50, then compare to library Storybook and verify that values match.

Export result to an XLSX and verify matching

## Checklist ##


components/containers/ColorFilterTool/ColorFilterTool.tsx
```diff
@@ -23,7 +23,7 @@ export const Fragment = graphql(`
     displayName
     filterToolActions
     filterColorOptionsLabels: filterColorOptions(label: true)
-    filterColorOptionsValues: filterColorOptions(label: true)
+    filterColorOptionsValues: filterColorOptions(label: false)
```
alexgoff, line 26: this cause 2 hours of debugging!

blnkt (changes requested):
First off, I rebased this PR on top of `develop` in anticipation of merging it in, so I'd advise `git reset --hard origin/EPO-8855` when you bring this branch up on your local.

- Color filter values in dropdown are correctly translated 
- Tested image generation (made an image in the color filter tool and then exported it in the xlsx) and it all seems to match now.  However, I did find that if `None` is selected for the color filters in the dropdowns the generated image looks correct on the page where I made it, but not on the review page or in the xlsx (see screenshots below).  Not sure if this is due to recent changes or is longstanding bug that went unnoticed up until now.  If the latter, add a new ticket to your backlog and can address outside of this PR.  
Screen Shot 2024-01-08 at 11 14 43 AM
Screen Shot 2024-01-08 at 11 14 24 AM
Screen Shot 2024-01-08 at 11 05 29 AM
 - I encountered an error when loading the review page in Spanish (see screenshot below).  Not sure if this is due to recent changes or is longstanding bug that went unnoticed up until now.  If the latter, add a new ticket to your backlog and can address outside of this PR.  
Screen Shot 2024-01-08 at 11 03 55 AM

blnkt (approved):
My previous comments were all addressed

PR #122, opened 2024-01-09
[EPO-8949] Fix Google OAuth

blnkt:
JIRA: https://jira.lsstcorp.org/browse/EPO-8949
## What this change does ##

The custom Google Login button is back!  With it we are using the [Auth Code flow](https://developers.google.com/identity/oauth2/web/guides/how-user-authz-works#oauth_20_flows) wherein the client receives an auth `code` from Google after the user auths, serverside we use a Google OAuthClient (part of the Google js SDK) to send this code to a Google OAuth endpoint, along with our `GOOGLE_APP_ID` and `GOOGLE_APP_SECRET`, and Google sends us back the `idToken` (jwt) we need to lookup/auth the user against Craft Users (via the GraphQL Authentication Craft plugin).

## Notes for reviewers ##

This PR is best tested alongside https://github.com/lsst-epo/investigations-client/pull/122.
Run yarn to pick up the newly added Google Auth package.

While it turned out the `GOOGLE_APP_SECRET` env var was superflous in the api, it is now very much needed in the client app in order to exchange the auth `code` for jwt `tokens`.  This exchange happens via server action called from a client component so I think this secret is used in a relatively secure way (i.e. not exposed), judging from [the next.js documentation](https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser)...I think.  Ask @blnkt if you're not sure what value to use for the `GOOGLE_APP_SECRET` env var.

## Testing ##

This PR is best tested alongside [this PR on the backend](https://github.com/lsst-epo/investigations-api/pull/62).  
Using the frontend UI at http://localhost:3000/coloring0the-universe, click the login button, select Educator group, and then click the "continue with Google" button to sign up as a new User.  This should open a Google content screen in a new tab, and/or prompt you to login to a Google Account.  After authing the tab should close itself and return you to the page you initiated the auth from.  It should also immediately log you in (i.e. no pending state; no prompt to verify your email before continuing).  Logout, and then log back in.  Please repeat this test with a student-type user too.  You may need to delete your new Educator type user first, as well as clear out [the permissions granted to our Google App in your Google Account settings](https://myaccount.google.com/connections?continue=https%3A%2F%2Fmyaccount.google.com%2Fsecurity).  Also observe that the button text on the "Google Login" changes, and that the button is disabled after sign up/in is initiated, remaining so until the user has been successfully authed (craft returns the tokens we need, we've saved them as cookies, and the current path has been revalidated).

N.B. Your lsst.org Google Workspace account does not count as a legit Google Identity to auth with because it has no gmail email address associated with it, and that and the user's name are the principal pieces of UserInfo we need to successfully auth.  Your NOIRLab email will work.

### DIFF
--- a/components/auth/buttons/GoogleSSO/Button.tsx
+++ b/components/auth/buttons/GoogleSSO/Button.tsx
@@ -1,73 +1,59 @@
 "use client";
 
-import type { GoogleCredentialResponse } from "@react-oauth/google";
-import { Button } from "@rubin-epo/epo-react-lib";
+import { useState } from "react";
 import { useTranslation } from "react-i18next";
+import { useGoogleLogin } from "@react-oauth/google";
+import { Button } from "@rubin-epo/epo-react-lib";
 import { useAuthDialogManager } from "@/contexts/AuthDialogManager";
 import { usePathToRevalidate } from "@/components/auth/clientHelpers";
-import { authenticateEducator, authenticateStudent } from "./actions";
-import { GoogleLogin } from "@react-oauth/google";
-// import { useGoogleLogin } from "@react-oauth/google";
+import { getTokens, authenticateUser } from "./actions";
 
 export default function SSOButton({
   onError: onSignInError,
-  buttonWidth
 }: {
   onError: () => void;
-  buttonWidth: number;
 }) {
-  const {
-    // t,
-    i18n: { language },
-  } = useTranslation();
+  const { t } = useTranslation();
   const { pendingGroup } = useAuthDialogManager();
   const pathToRevalidate = usePathToRevalidate();
-  // const login = useGoogleLogin({
-  //   onSuccess: handleSuccess,
-  //   onError: handleError,
-  // });
+  const login = useGoogleLogin({
+    onSuccess: handleSuccess,
+    onError: handleError,
+    flow: "auth-code",
+  });
+  const [isLoading, setIsLoading] = useState(false);
 
   function handleError() {
-    onSignInError(response);
+    onSignInError();
   }
 
-  async function handleSuccess(credentialResponse: GoogleCredentialResponse) {
-    try {
-      if (!credentialResponse.credential) return handleError();
+  async function handleSuccess(res) {
+    const { code } = res;
+    const { id_token: idToken } = await getTokens(code);
 
-      if (pendingGroup === "educators") {
-        await authenticateEducator(
-          credentialResponse.credential,
-          pathToRevalidate
-        );
-      } else {
-        await authenticateStudent(
-          credentialResponse.credential,
-          pathToRevalidate
-        );
-      }
+    try {
+      if (!idToken) return handleError();
+      await authenticateUser(pendingGroup, idToken, pathToRevalidate);
     } catch (error) {
-      onSignInError();
+      handleError();
+      setIsLoading(false);
     }
   }
 
-  // return (
-  //   <Button onClick={() => login()} styleAs="tertiary">
-  //     {t("sign_in.continue_with_google")}
-  //   </Button>
-  // );
   return (
-    <GoogleLogin
-      onSuccess={handleSuccess}
-      onError={handleError}
-      locale={language}
-      shape="rectangle"
-      size="large"
-      text="continue_with"
-      theme="outline"
-      type="standard"
-      width={buttonWidth}
-    />
+    <Button
+      onClick={() => {
+        setIsLoading(true);
+        login();
+      }}
+      styleAs="tertiary"
+      disabled={isLoading}
+      isBlock
+    >
+      {isLoading
+        ? t("sign_in.loading_google")
+        : t("sign_in.continue_with_google")}
+    </Button>
   );
 }
 
--- a/components/auth/buttons/GoogleSSO/GoogleSSO.tsx
+++ b/components/auth/buttons/GoogleSSO/GoogleSSO.tsx
@@ -1,8 +1,6 @@
 "use client";
 
-import { useState } from "react";
 import { GoogleOAuthProvider } from "@react-oauth/google";
-import useResizeObserver from "use-resize-observer";
 import Button from "./Button";
 
 const GOOGLE_APP_ID = process.env.NEXT_PUBLIC_GOOGLE_APP_ID;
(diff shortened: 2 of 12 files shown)


components/auth/buttons/GoogleSSO/Button.tsx
```diff
@@ -1,73 +1,58 @@
...
-    onSignInError(response);
+    onSignInError();
   }
 
-  async function handleSuccess(credentialResponse: GoogleCredentialResponse) {
-    try {
-      if (!credentialResponse.credential) return handleError();
+  async function handleSuccess(res) {
+    const { code } = res;
+    const { id_token: idToken } = await getTokens(code);
```
blnkt, line 32: I wasn't totally sure what this Type was, and whether or not the new Auth Code flwo response it was getting was indeed a `GoogleCredentialResponse` so I just chucked it.

components/auth/buttons/GoogleSSO/Button.tsx
```diff
@@ -1,73 +1,58 @@
...
-        );
-      } else {
-        await authenticateStudent(
-          credentialResponse.credential,
-          pathToRevalidate
-        );
-      }
+    try {
+      if (!idToken) return handleError();
+      await authenticateUser(pendingGroup, idToken, pathToRevalidate);
```
blnkt, line 36: new method that attempts to auth the user as an educator and a student since we don't know which one they actually are until after they are authed.

components/auth/buttons/GoogleSSO/Button.tsx
```diff
@@ -1,73 +1,58 @@
...
+      onClick={() => {
+        setIsLoading(true);
+        login();
+      }}
+      styleAs="tertiary"
+      disabled={isLoading}
+    >
+      {isLoading
+        ? t("sign_in.loading_google")
+        : t("sign_in.continue_with_google")}
```
blnkt, line 54: useingf `isLoading` state to disable the button and change the button text why auth is happening.

components/auth/buttons/GoogleSSO/GoogleSSO.tsx
```diff
@@ -14,25 +12,11 @@ export default function GoogleSSO({
...
   onError: () => void;
 }) {
-  const [buttonWidth, setButtonWidth] = useState();
-
-  const { ref } = useResizeObserver({
-    onResize: ({ width }) => {
-      setButtonWidth(width);
-    },
-  });
-
```
blnkt, line 24: Removed all this junky stuff we needed for the NOT custom button.

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -1,11 +1,28 @@
...
+  "postmessage"
+);
+
+// exchanges Auth Code for Tokens
+export async function getTokens(code) {
+  const { tokens } = await oAuth2Client.getToken(code);
+
+  return tokens;
+}
+
```
blnkt, line 25: new method that exchanges the auth `code` for `tokens`

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -14,6 +31,17 @@ const StudentMutation = graphql(`
...
+  if (authData) {
+    setAuthCookies(authData);
+
+    if (pathToRevalidate) {
+      revalidatePath(pathToRevalidate, "page");
+    }
+    return authData;
+  }
+}
+
```
blnkt, line 44: abstracted this out to reuse across the other methods in this file

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -58,22 +80,60 @@ export async function authenticateEducator(
...
+  idToken: string,
+  pathToRevalidate?: string
+) {
+  // There might be a case where the group is already set, so we check first
+  if (group === "educators") {
+    await authenticateEducator(idToken, pathToRevalidate);
+  } else if (group === "students") {
+    await authenticateStudent(idToken, pathToRevalidate);
+    // Usually group is not set, so we'll try pulling user data from both students and educators
+  } else {
```
blnkt, line 107: Not sure the pendingGroup will ever have value when `authenticateUser` is called, but felt like didn't hurt anything to check.

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -58,22 +80,60 @@ export async function authenticateEducator(
...
+      variables: { idToken },
+    });
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const studentAuthData = useFragment(
+      AuthFragmentFragmentDoc,
+      studentData?.googleSignInEducators
+    );
+    conditionallySetCookiesRevalidatePath(studentAuthData, pathToRevalidate);
+    // Returns if data so we don't hit the educator data fetch if we don't need it
+    if (studentAuthData) return studentAuthData;
```
blnkt, line 120: Try authing as a student and return something so we don't try to auth as an educator if we don't need to.

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -58,22 +80,60 @@ export async function authenticateEducator(
...
+      query: EducatorMutation,
+      variables: { idToken },
+    });
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const eduAuthData = useFragment(
+      AuthFragmentFragmentDoc,
+      eduData?.googleSignInEducators
+    );
+    conditionallySetCookiesRevalidatePath(eduAuthData, pathToRevalidate);
+    if (eduAuthData) return eduAuthData;
```
blnkt, line 133: Try authing as an educator

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -58,22 +80,60 @@ export async function authenticateEducator(
...
+
+    if (eduError || studentError) {
+      throw new Error(eduError?.message || studentError?.message);
     }
```
blnkt, line 137: If none of the above worked, and there is an error, throw it!

contexts/AuthDialogManager.tsx
```diff
@@ -49,16 +49,16 @@ function AuthDialogManagerProvider({
     getInitActive(searchParams)
   );
 
-  const [pendingGroup, setPendingGroup] = useState<PendingGroup>("students");
+  const [pendingGroup, setPendingGroup] = useState<PendingGroup>(null);
```
blnkt, line 52: For the above to all work right we need the `pendingGroup` to NOT have a default value.

package.json
```diff
@@ -66,6 +66,7 @@
     "exceljs": "^4.4.0",
     "focus-trap": "^7.3.1",
     "focus-visible": "^5.2.0",
+    "google-auth-library": "^9.4.1",
```
blnkt, line 69: added new package for auth code flow stuff

alexgoff:
Received a bad request when signing up as an Educator using my NoirLab email which is already registered to my admin account: 

image

PR #127, opened 2024-01-16
[F] video content block

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8890

## What this change does ##

Client branch for video content blocks. Mostly based on Image blocks, this branch takes some functions from that component and formalizes them for use in other asset fields.

## Notes for reviewers ##

Check changes on both the client and API. Changes can be checked by adding a Video content block to any page, select a DAM video asset and insert into the page. View your page on the client and see that it correctly attached the URL and shows in a player. 

- 4

## Checklist ##


components/factories/ContentBlockFactory/ContentBlockFactory.tsx
```diff
@@ -40,43 +45,26 @@ const Fragment = graphql(`
     ...ReferenceModalBlock
     ...ColorFilterToolBlock
     ...EquationBlock
+    ...VideoBlock
   }
 `);
 
-interface ContentBlockFactoryProps {
+interface ContentBlockFactoryProps extends BaseContentBlockProps {
```
alexgoff, line 52: did a little bit of cleanup here to make sure the props we need are always passed down to the final block, and establish unify the props between the factory and final block

components/content-blocks/Video/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+      metadata: additional {
+        CaptionEN
+        CaptionES
+        Credit
+      }
+    }
+  }
+`);
+
+const VideoBlock: FunctionComponent<
```
alexgoff, line 29: instead of redeclaring a props interface just to overwrite the `data` property with the `FragmentType<typeof Fragment>` we can do that using generic types on `BaseContentBlockProps`

blnkt (approved):
Nice work.

PR #134, opened 2024-02-08
[F] Interaction text on widget modals

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8998

## What this change does ##

Adds the button that displays the interaction instructions for widgets while they are open in modals. Also took the opportunity to clean up the widget container styling and try to reduce the number of styled overrides down into a couple of defined states (light theme, dark theme, non-interactive, filled screen, etc.)

## Notes for reviewers ##

API + client change. API change is just to add widget instructions as a field on static widgets. 

Test by first viewing any widgets that are associated with questions 

- 2

## Testing ##

Can test using question widgets (Color Tool) by opening the widget in it's modal and observing that the "info" button is present at the bottom of the modal.

Can test in static widgets (Filter Tool, Camera Filter Tool) by adding instructional text (delete the existing text block above it) and observing the same.

## Checklist ##


components/content-blocks/ColorFilterTool/styles.ts
alexgoff, line 1: tried to remove as many custom styled widget containers as possible and create standard variants

components/layout/WidgetContainer/WidgetContainer.tsx
```diff
@@ -19,31 +24,31 @@ const padding = {
...
   className,
   style,
 }) => {
   return (
     <Styled.WidgetContainer
       data-modal-open={isOpen}
+      data-theme={variant}
+      data-interactive={interactive}
+      data-fill-screen={fillScreen}
       className={className}
```
alexgoff, line 50: tried to reduce the widget containers down to a few states:

- opened or closed modal
- light or dark theme (ex. color tool is light, prism filter tool is dark)
- if the widget should try to fill screens on small sizes (ex. saved image color filter tool does not try to fill the screen because it is more like an image, but the editable tool does)

PR #136, opened 2024-02-12
[F] Supernova 3 Vector Widget

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-9037

## What this change does ##

Adds Supernova 3 Vector widget to the client

## Notes for reviewers ##

Client and API change, grab EPO-8998 for the API as well. 

- 4

## Testing ##

**API**

- Go to an investigation page
- Add "Supernova Three Vector" widget
- Add configuration for it. _I debated between making this an entry widget or a content block widget. It is not part of a question so it doesn't really fit the entries, but it requires much more config than other content block widgets. I think this is ultimately the right place for it, but open to input._

**Client**

- Go to page with the widget set up
- Observe that there is a brief loader before it appears, this is because the widget is dynamically loaded client-side only. 
- Test the same functionality that was in the library, should be no major difference apart from opening it in the modal.

## Checklist ##


components/content-blocks/InteractionGroupContainer/InteractionGroupContainer.tsx
```diff
@@ -13,11 +13,10 @@ const Fragment = graphql(`
       ...ImageBlock
       ...TableBlock
       ...QuestionsBlock
-      ...BarGraphToolBlock
       ...FilterToolBlock
-      ...ScatterplotToolBlock
       ...ReferenceModalBlock
       ...ColorFilterToolBlock
+      ...SupernovaThreeVectorBlock
     }
```
alexgoff, line 20: removed blocks that were created in early development and never used

components/content-blocks/SupernovaThreeVector/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  if (!url) return null;
+
+  const response = await fetch(url, {
+    cache: "force-cache",
+    headers: { "Content-Type": "application/json" },
+    next: { tags: ["datasets"] },
+  });
+  const supernovaData: Array<number> = await response.json();
+
+  return (
```
alexgoff, line 57: the content block does all of the on-server data retrieval and set, then passes it off to the dynamic container to finish setup.

components/hoc/withModal/withModal.tsx
```diff
@@ -10,7 +10,7 @@ import * as Modal from "@/layout/Modal";
  * pulled out and shown in a modal, rather than off-page content
  * that is not visible until opened.
  */
-function withModal<T extends BaseContentBlockProps>(
+function withModal<T extends ModalProps>(
```
alexgoff, line 13: extracted out the modal specific props from the content block props because not all content blocks will have modal props and not all modals will have content block props (like the supernova three vector modal which is several steps removed from being a content block)

blnkt:
out of sync with the backend until after the 3 vector work is redone on the backend.  keeping PR open for reference until that work is done.

PR #139, opened 2024-02-21
[F] Supernova Distance & Distribution

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-9037

## What this change does ##

Redo of Supernova Distance & Distribution widget

## Notes for reviewers ##

Review on frontend and client

- 5 (mostly because of everything leading up to this)

## Testing ##

Add a Supernova Distance & Distribution widget to a page, and check there.

- Widget should appear in a widget container
- Widget should open into a modal
- Widget should include instructions (also visible in the modal)

## Checklist ##


blnkt (changes requested):
- [x] safeties to handle cases where user makes "bad" selections in Craft for Image Album or histogramData
- [x] Bin should be 100 by default
- [ ] Slider should slide in fullscreen mode

PR #143, opened 2024-03-01
[B] studentAuthData has student object for useFragment arg

blnkt:
To be reviewed alongside https://github.com/lsst-epo/investigations-api/pull/80

Please test:

- user creation with google auth and with username + password
- user login with google auth and with username + password

Note for reviewers:
 - In my testing the forget password link and the account activation emails are not sending correctly.  I am willing to believe this is a problem unique to local dev and so suggest testing those additional use cases after this work has been released to `dev`.
 - You will need an env var for the google secret in addition to the app id for the new client auth code to to work as expected.  @alexgoff you may already have those in there from my previous refactor of this google auth work, but if you need either ping @blnkt in Slack.

### DIFF
--- a/components/auth/buttons/GoogleSSO/actions.ts
+++ b/components/auth/buttons/GoogleSSO/actions.ts
@@ -19,7 +19,6 @@ const oAuth2Client = new OAuth2Client(
 // exchanges Auth Code for Tokens
 export async function getTokens(code) {
   const { tokens } = await oAuth2Client.getToken(code);
-
   return tokens;
 }
 
@@ -80,12 +79,14 @@ export async function authenticateEducator(
     query: EducatorMutation,
     variables: { idToken },
   });
+
   // not actually a client-side hook, just named like one
   // eslint-disable-next-line react-hooks/rules-of-hooks
   const authData = useFragment(
     AuthFragmentFragmentDoc,
     data?.googleSignInEducators
   );
+
   conditionallySetCookiesRevalidatePath(authData, pathToRevalidate, error);
 
   if (error) {
@@ -113,7 +114,7 @@ export async function authenticateUser(
     // eslint-disable-next-line react-hooks/rules-of-hooks
     const studentAuthData = useFragment(
       AuthFragmentFragmentDoc,
-      studentData?.googleSignInEducators
+      studentData?.googleSignInStudents
     );
     conditionallySetCookiesRevalidatePath(studentAuthData, pathToRevalidate);
     // Returns if data so we don't hit the educator data fetch if we don't need it
--- a/components/auth/buttons/GoogleSSO/actions.ts
+++ b/components/auth/buttons/GoogleSSO/actions.ts
@@ -10,6 +10,28 @@ import { AuthFragmentFragmentDoc } from "gql/public-schema/graphql";
 const GOOGLE_APP_ID = process.env.NEXT_PUBLIC_GOOGLE_APP_ID;
 const GOOGLE_APP_SECRET = process.env.GOOGLE_APP_SECRET;
 
+const StudentMutation = graphql(`
+  mutation GoogleSignInStudent($idToken: String!) {
+    googleSignInStudents(idToken: $idToken) {
+      ...AuthFragment
+    }
+  }
+`);
+
+const EducatorMutation = graphql(`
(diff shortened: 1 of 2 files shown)


components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -113,7 +114,7 @@ export async function authenticateUser(
     // eslint-disable-next-line react-hooks/rules-of-hooks
     const studentAuthData = useFragment(
       AuthFragmentFragmentDoc,
-      studentData?.googleSignInEducators
+      studentData?.googleSignInStudents
```
blnkt, line 117: typo in the key meant this was always failing when user was a student.  Because the path only revalidates if the data was made, on the frontend this would make it appear as if the user had not successfully completed a sign up flow, even when the user record was being created in the backend.

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -10,6 +10,28 @@ import { AuthFragmentFragmentDoc } from "gql/public-schema/graphql";
...
+    googleSignInEducators(idToken: $idToken) {
+      ...AuthFragment
+    }
+  }
+`);
+
+const userGroups = {
+  students: { mutation: StudentMutation, directive: "googleSignInStudents" },
+  educators: { mutation: EducatorMutation, directive: "googleSignInEducators" },
+};
```
blnkt, line 32: A lil dictionary to best facilitate the refactoring (combining) of educator & student auth methods

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -42,54 +55,25 @@ function conditionallySetCookiesRevalidatePath(authData, pathToRevalidate) {
...
-  idToken: string,
-  pathToRevalidate?: string
-) {
-  const { data, error } = await mutateAPI({
-    query: EducatorMutation,
-    variables: { idToken },
-  });
-  // not actually a client-side hook, just named like one
-  // eslint-disable-next-line react-hooks/rules-of-hooks
-  const authData = useFragment(
```
blnkt, line 85: This file looks messier than it is because I not only refactored work, I also reordered the methods.  Sufficed to say, the big thing I did here was I deleted the dedicated `authenticateStudent` and `authenticateEducator` methods, created a new generic `autheticate` method that takes a group (`students` or `educators`) as an arg and therefore can auth either group.

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -42,54 +55,25 @@ function conditionallySetCookiesRevalidatePath(authData, pathToRevalidate) {
...
-  );
-  conditionallySetCookiesRevalidatePath(authData, pathToRevalidate, error);
-
-  if (error) {
-    throw new Error(error.message);
+function conditionallyHandleError(data, error) {
+  if (!data && error) {
+    throw new Error(error?.message);
+  } else if (error) {
+    console.warn(error?.message);
```
blnkt, line 76: Something I ran into over and over when troubleshooting/testing, is that the gql-auth plugin will propogate `[GraphQL] User doesn't have permission to perform this mutation` error even when the auth operation is "successful" (i.e. you get user data back).  The error seems to be erroneous for us then.  Therefore, we check for data, and if there is data, we just do a lil soft console.log warning instead of throwing an error/exception.

components/auth/buttons/GoogleSSO/actions.ts
```diff
@@ -98,42 +82,35 @@ export async function authenticateUser(
...
+  // If there's a group selected this is a sign up
+  if (group) {
+    const { authData, error } = await authenticate(
+      idToken,
+      group,
+      pathToRevalidate
+    );
+
+    conditionallySetCookiesRevalidatePath(authData);
+    conditionallyHandleError(authData, error);
```
blnkt, line 94: As far as I can tell the only case where there is a group is already set is when a user selected their group in the UI (i.e. on Sign Up), and so we have to honor that selection here, otherwise all new users become students regardless of their selection in the UI.  @alexgoff in my testing, and when digging into the code, I never saw an edge case where the group is set through a cookie.  There are cases where the group is read out of the cookie, but I believe only ever after a user is authed, and therefore not germane to this initial user flow.

alexgoff (approved):

PR #144, opened 2024-03-01
[F] Integrate light curve widgets

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-9061

## What this change does ##

Adds light curve widgets in 3 places:

- As an optional addon to source selectors
- As a static display of datasets
- As an interactive question for students to determine apparent magnitudes

## Notes for reviewers ##

Change is on client and API

- 6

## Testing ##

Test using this branch, and EPO-9061 on the API.

Source Selector add-on
- In Craft
    - On an existing Source Selector widget, check "Include scatter plot" and observe that configuration for the plot X/Y axis appears. No other change is necessary since the scatterplot uses the same dataset.
    - Ideally I wanted to make this a very generic approach of simply adding more widgets to the question and displaying them side by side. Unfortunately these widgets share a state (active alert) and I don't have a good way of doing that at the moment without a bespoke component. I think there is a case for future work on creating a sort of shared state between widget blocks on the same question like:
    
    ```jsx
    const [widgetState, setWidgetState] = useState({})

   return widgets.map(widget => <Widget state={widgetState} onStateChange={(v) => setWidgetState(v)} />)


but the scope of this change is already quite large and I don't want to creep further. 
- In client
    - Observe that the light curve scatter plot appears next to (or underneath, on mobile) the source selector with an empty plot. 
    - Click on the source in the source selector and observe that the plot appears.
    - If the source is selected, and the widget opened into fullscreen, observe that the light curve plot shows the apparent magnitude for the current alert shown on the source selector.
    - Review page should only show the selected source on a single static image (no blinking)
    - Review XLSX should only have the selected source name

Static widget

- In Craft
    - Under "Widgets" create a new Light Curve Tool and assign a dataset
    - On an Investigations page, add a widget content block and assign the Light Curve Tool that you created
- In client
    - Observe a static light curve scatter plot 

Interactive question

- In Craft
    - Under "Widgets" create a new Light Curve Tool and assign one or many datasets (this allows randomization of the data)
    - Under "Questions" create a new widget question and assign the Light Curve Tool you created to the widgets block
- In client
    - Observe that an interactive version of the light curve tool is shown, with controls to modify the curve and read off the deltam15 and apparent magnitude
    - The review page should show the student's work on the plot in a static form
    - The review XLSX should show the apparent magnitude from the slider on the plot

## Checklist ##

## Screenshots (optional) ##
### Before:
### After:


components/content-blocks/MagnitudeScatterPlot/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [{ peakMjd, json, title }] = dataset;
+
+  const alerts = await getDataset(json[0]?.url || undefined);
+
+  return (
+    <WidgetContainerWithModal
+      title={t("widgets.light_curve.title") || undefined}
+      interactive={false}
+    >
+      <ObservationsPlot
```
alexgoff, line 73: it would be nice to use the dynamic container I created here but since that does a clientside fetch I just did the same wrapper actions in the content block

lib/api/hooks/useAlerts.ts
```diff
@@ -0,0 +1,18 @@
...
+import useSWR from "swr";
+
+interface Alert {
+  id: number;
+  error: number;
+  date: number;
+  magnitude: number;
+}
+
+const useAlerts = (url?: string) => {
```
alexgoff, line 11: helpful wrapper for the `SWR` call to get the alert assets, does not make a call if a URL is not supplied, and will de-dupe requests to the same URL.

PR #146, opened 2024-03-06
[F] Tabular Questions

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-8938

## What this change does ##

Adds the following:

- Tabular questions
    - Tables can reference previous tabular questions
- Content block tables can reference previous tabular questions
- Tabular questions on review pages
- Tabular questions in XLSX exports

## Notes for reviewers ##

This PR is on the client and API, the main impact to the API is a "tabular" question type that can be created, and that "Display Tables" can reference tabular questions in their own table.

- 5

## Testing ##

**API**
- Under "Question" entries, create a new question and observe that "Tabular" is now available
- Create a tabular question and give it a header row, followed by some cells with inputs or static content.
    - You could, in theory, create an entirely static table underneath a question. I don't really see a purpose to doing that though, but also not really a reason to forbid it.
    - Observe that the question can be completed and it's entries saved
- Create another tabular question and reference your previous question within it.
    - Observe that the previous results are shown in your table alongside the new inputs
- Create a new content block table and reference your previous question within it.
    - Referenced questions will _only_ show their content and not any questions referenced within them

**Client**

- Observe that math equations render correctly in tables
- Observe that the question table shows on the review page
- Observe that the question table has a condensed version of itself on the printed review page
- Observe that the XLSX export outputs the entire table in a single cell, with each header value shown next to it's value for the row. Math equations are shown using a simplified interpretation (basically just HTML minus the tags) in this export.

**Additional**

The review page has been refactored to push server rendering out past the ReviewFactory so each review item can decide when to switch over to client rendering (to retrieve it's answer from the context). Do a double check that the review page is still working as intended.

## Checklist ##


app/[locale]/[investigation]/review/page.tsx
```diff
@@ -11,12 +11,53 @@ import {
...
 import ReviewTemplate from "@/components/templates/ReviewPage";
 import { getSite } from "@/helpers";
+import { getPageQuestionEntries } from "@/helpers/questions";
 
 const Query = graphql(`
   query ReviewPage($site: [String], $uri: [String]) {
     entry(site: $site, uri: $uri) {
       ... on investigations_investigationParent_Entry {
         title
+        children {
```
Comment (alexgoff) [inline app/[locale]/[investigation]/review/page.tsx:21]: unfortunately to power the review page we need to grab all of the questions from the server here

app/[locale]/[investigation]/review/page.tsx
```diff
@@ -46,8 +87,15 @@ const ReviewPage: FunctionComponent<ReviewPageProps> = async ({
     },
   });
 
+  const questions = data?.entry.children
```
Comment (alexgoff) [inline app/[locale]/[investigation]/review/page.tsx:90]: moved all of the code from the `Questions` context to a helpers file so they can be used server-side and get questions in the same format as the Questions context

app/api/xlsx/formatter/index.ts
```diff
@@ -53,21 +52,22 @@ export type InlineFormatterFactory = (
...
   | SelectFormatter
   | InlineFormatterFactory
-  | WidgetFormatterFactory;
+  | WidgetFormatterFactory
+  | TabularFormatter;
 
-const formatters: {
-  [key: string]: Formatter;
-} = {
+const formatters: Record<string, Formatter> = {
```
alexgoff, line 64: cleanup and add tabular as an XLSX formatter option

app/api/xlsx/formatter/tabular/index.ts
```diff
@@ -0,0 +1,50 @@
+import temml from "temml";
+import { serverTranslation } from "@/lib/i18n";
+import { TabularFormatter } from "..";
+import { Cell } from "@/components/questions/TabularQuestion/Table";
+
+const renderStaticContent = ({ text, equation }: Partial<Cell>) => {
+  if (equation) {
+    const mathml = temml.renderToString(equation).replace(/<[^>]*>/g, "");
```
alexgoff, line 8: do a super simple interpretation of the equation MathML output by removing the HTML tags and keeping the text and symbols

app/api/xlsx/formatter/tabular/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  cell,
+  locale,
+  rows,
+  value,
+}) => {
+  const { t } = await serverTranslation(locale, "translation");
+  const { cells: header = [] } = rows.shift() || {};
+  const headerLabels = header.map(renderStaticContent);
+
+  const parsedRows = rows.reduce(
```
alexgoff, line 26: this runs down each row and matches the cell with it's header to display data in a list format like:

```
Supernova: A
m15: 0.42
Apparent: 17
Absolute: 18
Distance: 320
```

components/atomic/Equation/index.tsx
```diff
@@ -0,0 +1,23 @@
+import "temml/dist/Temml-Local.css";
+import { FunctionComponent } from "react";
+import temml from "temml";
+import { LatinModernMath } from "@/lib/fonts";
+import * as Styled from "./styles";
+
+const Equation: FunctionComponent<{ latex: string }> = ({ latex }) => {
```
alexgoff, line 7: the Equation API is dead, long live the Equation component. The correct Next approach for hefty server libraries is to simply render in a server component rather than making an API call from the server. This feature was a good opportunity since the table cells are rendering in multiple loops so awaiting an internal API call is a useless latency.

This Equation server component does so and has the nice benefit of always importing the CSS and font library when in use. Equation API should eventually be deprecated but it's fine for now.

components/content-blocks/Questions/index.tsx
```diff
@@ -33,7 +33,7 @@ const QuestionsContentBlock: FunctionComponent<
         {!!data.questionEntries?.length &&
           data.questionEntries.map((question, i) => {
             return question?.__typename === "questions_default_Entry" ? (
-              <QuestionFactory key={i} data={question} />
+              <QuestionFactory key={i} data={question} locale={locale} />
```
alexgoff, line 36: need to pass down locale so server rendered questions like tables can use it

components/content-blocks/Table/Table.tsx
```diff
@@ -1,61 +1,75 @@
...
 import * as Styled from "./styles";
+import { buildHeader, buildRows } from "@/components/layout/Table/helpers";
 
 const Fragment = graphql(`
   fragment TableBlock on contentBlocks_table_BlockType {
     id
     caption
     contentHeading
-    displayTable {
+    rows: displayTable {
```
alexgoff, line 14: added previous questions to the table content block and aliased some of the properties to line up with the table helpers developed for tabular questions

components/factories/ReviewFactory/index.tsx
```diff
@@ -1,12 +1,10 @@
 import { ComponentType, FunctionComponent } from "react";
 import * as Review from "@/components/questions/Review";
-import { AnswerData } from "@/types/answers";
 
 export interface ReviewProps {
   id: string;
   type: string;
   number: number;
-  value?: AnswerData;
```
alexgoff, line 9: answer data is no longer passed down, review items must retrieve their own answers with `useAnswer`

components/form/Input/styles.ts
```diff
@@ -11,12 +11,12 @@ export const Input = styled.input`
...
   color: var(--input-color);
   display: block;
-  font-size: 1rem;
+  font-size: 1em;
   margin-block-start: 1em;
   padding: 0;
   padding-inline: 1ch;
   width: 100%;
-  height: 2rem;
+  height: 2em;
```
alexgoff, line 19: should be responsive to font sizing

components/layout/Table/helpers.tsx
```diff
@@ -0,0 +1,84 @@
...
+    return {
+      id: id || undefined,
+      children: <Question {...{ id, questionId, readOnly, options }} />,
+    };
+  }
+
+  return buildStaticCell(cell);
+};
+
+export const buildRows = (
```
alexgoff, line 64: builder methods that are used among all versions of the table

components/layout/Table/index.tsx
```diff
@@ -37,35 +40,59 @@ const Table: FunctionComponent<TableProps> = ({
...
+          }
+
+          setCellIndex(cellIndex);
+
+          return;
+        }
+      }
+    }
+  };
+  const scrollBackward = () => {
```
alexgoff, line 79: split the scroller methods into two for easier debugging

components/layout/Table/styles.ts
```diff
@@ -9,7 +9,11 @@ export const TableWrapper = styled.div`
 
   padding-inline: var(--table-padding, 0);
   position: relative;
-  overflow: hidden;
+
+  @media only print {
```
alexgoff, line 13: some print styles for the review page

components/questions/Review/Inline/index.tsx
```diff
@@ -27,10 +29,12 @@ const REVIEW_MAP: Record<InlineQuestionType, ComponentType<any>> = {
 };
 
 const InlineReview: FunctionComponent<InlineReviewProps> = ({
+  id,
   number,
   parts = [],
-  value: partsValues = {},
 }) => {
+  const { answer = {} } = useAnswer<InlineQuestionData>(id);
```
alexgoff, line 36: refactor to pull it's own answer instead of receiving as a prop (same with the following)

components/questions/TabularQuestion/ReadOnly/index.tsx
```diff
@@ -0,0 +1,8 @@
+import { FunctionComponent } from "react";
+import * as Styled from "./styles";
+
+const ReadOnlyCell: FunctionComponent<{ value?: string }> = ({ value }) => {
```
alexgoff, line 4: readonly input for displaying previous answers

components/questions/TabularQuestion/Table/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+  __typename: string;
+  id: string;
+  equation?: string;
+  text?: string;
+  questionType?: "text" | "select";
+  header?: boolean;
+  options?: Array<Option>;
+}
+
+const getOverflowPadding = (
```
alexgoff, line 86: method that figures out how much padding to add to the bottom of the table for select dropdowns so there is not a y overflow that breaks the table

components/review/Export/index.tsx
alexgoff, line 1: moved a bunch of review page stuff into their own section in components

components/templates/ReviewPage/index.tsx
```diff
@@ -1,63 +1,37 @@
...
-        </time>
-      </div>
-      <ReviewList {...{ answers, questions }} />
-      <FormButtons>
-        <PrintButton filename={name}>{t("review.print")}</PrintButton>
-        <XlsxButton {...{ answers, questions, investigation, name }}>
-          {t("review.export")}
-        </XlsxButton>
-      </FormButtons>
+      <ReviewExport {...{ user, questions, investigation }}>
```
alexgoff, line 22: the big change in server rendering this page is that all of the interactive elements (name, print, export) are put into their own component so the review list itself can stay on the server

helpers/questions.ts
alexgoff, line 1: all of the methods that were in the question context are in here now, so the server rendered review page can get them

lib/utils.ts
```diff
@@ -135,3 +135,11 @@ export function normalizeItemData(items: any[], whichEntry = "entry") {
 
 export const getDisplayName = (component: ComponentType<any>) =>
   component.displayName || component.name || "Component";
+
+export function notNull<T>(value: T): value is NonNullable<T> {
+  return !isNullish(value);
+}
+
+export const isNullish = <T>(
```
alexgoff, line 143: finally some null utilities to deal with codegen's annoying addition of `| null` to every type

PR #149, opened 2024-03-11
[F] Add calculator questions

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-9041

## What this change does ##

Adds calculator type questions, starting with the Absolute Magnitude and Supernova Distance equations. Includes interactive (question), static (review), and XLSX variants. Some changes to the general Temml setup for better font support.

## Notes for reviewers ##

Use this branch and corresponding API. Most of the changes will be reviewed on the frontend.

- 4

## Testing ##

- API:
    - Create a new question with the "Calculator" type and select an equation to use. Add these questions to a page.
- Client:
    - Question:
        - See that your calculator appears beneath it's text, with formatted math, serif font for the non-interactive parts, and sans serif for the inputs.
        - See that there is no value shown in the output, if no values have been input.
        - See that a value is shown if the value in the input is zero or another real number.
    - Review:
        - See that a static rendering of the equation is shown in similar format to the question 
    - XLSX:
        - See that the result of the question is added to it's table cell

## Checklist ##


components/calculators/interactive/peakAbsoluteMagnitude.tsx
```diff
@@ -0,0 +1,72 @@
+import { ChangeEvent, FunctionComponent } from "react";
+import { useTranslation } from "react-i18next";
+import { InteractiveCalculatorProps } from "@/types/calculators";
+import MathInput from "@/components/form/Input/patterns/MathInput";
+import Output from "../Output";
+
+const PeakAbsoluteMagnitude: FunctionComponent<
```
alexgoff, line 7: you'll notice that these components have a Typescript error that the MathML elements are not part of JSX typings, this is true although it doesn't impact rendering since React recognizes them fine.

Ideally I would like to do this all with LaTeX to MathML and avoid writing any MathML, but I could not find any renderers that support injecting React components for the inputs to work.

components/calculators/math/index.ts
```diff
@@ -0,0 +1,52 @@
+import { Calculator, Equation } from "@/types/calculators";
+import round from "lodash/round";
+
+const peakAbsoluteMagnitude: Calculator = ({ m15 }) => {
```
alexgoff, line 4: each equation's math and an output in LaTeX format is stored in these calculator methods

components/calculators/static/index.tsx
```diff
@@ -0,0 +1,19 @@
+import { FunctionComponent } from "react";
+import { CalculatorValues, Equation } from "@/types/calculators";
+import Equations from "@/components/calculators/math";
+import EquationRenderer from "@/components/atomic/Equation";
+
+const StaticCalculator: FunctionComponent<{
```
alexgoff, line 6: static version of the calculators that simply grabs the LaTeX output

components/content-blocks/Equation/index.tsx
```diff
@@ -21,18 +19,7 @@ const EquationContentBlock: FunctionComponent<TextProps> = async (props) => {
...
-
-  const result = await fetch(uri).then((response) => response.json());
-
-  return (
-    <Styled.MathWrapper
-      className={LatinModernMath.variable}
-      dangerouslySetInnerHTML={{ __html: result }}
-    ></Styled.MathWrapper>
-  );
+  return <Equation {...{ latex }} />;
```
alexgoff, line 22: consolidate existing MathML outputs into the `Equation` component

components/factories/QuestionFactory/index.tsx
```diff
@@ -3,22 +3,24 @@ import { graphql, useFragment, FragmentType } from "@/gql/public-schema";
...
+import CalculatorQuestion from "@/components/questions/Calculator";
 
 const Fragment = graphql(`
   fragment QuestionEntry on questions_default_Entry {
     __typename
     id
     answerType
+    questionText
+    widgetInstructions
+    ...CalculatorQuestion
```
alexgoff, line 16: moving toward getting all questions into their own fragments. At some point in the future I'd like to break out all of the inline questions, "simple" questions, and widget questions into their own files with own fragments. The existing setup is ok but it was developed before we went all in on fragment typings

components/questions/QuestionNumber/styles.ts
```diff
@@ -0,0 +1,15 @@
+"use client";
+import styled from "styled-components";
+
+export const Number = styled.li`
```
alexgoff, line 4: extract a bunch of the styles that were being implemented repeatedly across multiple questions

alexgoff:
Unfortunately running into an issue that I didn't pick up on during development, but now can't seem to resolve:

- When navigating to a page from another page in the application, the input boxes inside of the equations are non-functional. 
    - What is happening is that the inputs are not getting parsed as HTML. They are nested in MathML which is valid since they are in token elements (mi, mn, mo, ms, mspace, mtext) but the browser seems to disagree (sometimes)
    - Most of the functions of `input` elements are broken (value, placeholder, onChange) and the element is treated like a non-editable `generic` role when this happens.
    - This seems to be present across all browsers, Safari is the worst case and nothing renders when going backward/forward to the page.
    - This does not happen when fresh loading a page and fixes when reloading a broken page. 

Solutions I am thinking through right now:

- keep debugging and figure out how to force the element to be parsed correctly
- figure out an alternate way to render equations that uses styled HTML elements
- move the inputs outside of the equation (requires redesign)

Correct
image

Incorrect
image

blnkt (approved):

PR #154, opened 2024-04-01
[B] fix select dropdown behind question text

alexgoff:
For JIRA: "EPO-9125 #IN-REVIEW #comment fix select dropdown behind question text"

JIRA: https://jira.lsstcorp.org/browse/<ID>

## What this change does ##

Select dropdowns are already given their own stacking context, however since the browser considers items added later in the DOM to be at a higher implicit z-index they end up rendering behind dropdowns in following questions.

To fix this, the questions index among all questions is used to assign a `z-index` to that specific question, in descending order (first question will have the highest z-index)

## Notes for reviewers ##

Change is clientside

- 2

## Testing ##

Verified locally on Chrome, Safari, and Firefox. 

- Go to a page with multiple select questions (pages 6 and 10 in Coloring the Universe)
   - page 18 for additional verification, since it has the only non-inline selects that overlap
- Open select dropdowns, verify they do not overlap

## Checklist ##

### Before:
image

### After:
image

### DIFF
--- a/components/content-blocks/Questions/styles.ts
+++ b/components/content-blocks/Questions/styles.ts
@@ -13,6 +13,7 @@ export const QuestionList = styled.ol`
 
   background-color: var(--list-background-color, transparent);
   color: var(--question-color);
+  isolation: isolate;
   list-style-type: decimal;
   list-style-position: inside;
   padding: var(--list-padding, 0);
--- a/components/questions/QuestionNumber/styles.ts
+++ b/components/questions/QuestionNumber/styles.ts
@@ -3,6 +3,7 @@ import styled from "styled-components";
 
 export const Number = styled.li`
   container-type: inline-size;
+  position: relative;
 
   &[data-direction="block"] {
     & > * + * {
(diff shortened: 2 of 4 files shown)


components/content-blocks/Questions/styles.ts
```diff
@@ -13,6 +13,7 @@ export const QuestionList = styled.ol`
 
   background-color: var(--list-background-color, transparent);
   color: var(--question-color);
+  isolation: isolate;
```
alexgoff, line 16: questions list is given a new stacking order so the questions don't render in front of the pager or anything else

components/questions/QuestionNumber/styles.ts
```diff
@@ -3,6 +3,7 @@ import styled from "styled-components";
 
 export const Number = styled.li`
   container-type: inline-size;
+  position: relative;
```
alexgoff, line 6: make each question z-indexable

components/content-blocks/Questions/styles.ts
blnkt: never have I ever encountered the `isolation` property.  Does seem to do the trick

blnkt (approved):

PR #157, opened 2024-04-02
[C] refactor calculators

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-9131

## What this change does ##

- keep only vertical layout
- limit supernova distance to 3 significant figures
- heavy re-write to remove per-component calculators, try to use generic's to render calculators so that future calculators are simpler to plug in vs. writing a new component

## Notes for reviewers ##

clientside only

- 3

## Testing ##

- Add both types of calculator to a page
- Verify that the vertical layout looks ok across screen sizes
- Verify that the output of supernova distance is limited to 3 significant figures
- Verify that existing functionality is maintained

## Checklist ##


components/calculators/interactive/index.tsx
```diff
@@ -1,14 +1,61 @@
...
+/* eslint-disable jsx-a11y/label-has-associated-control */
+import { FormEvent, FunctionComponent } from "react";
+import { useTranslation } from "react-i18next";
+import { InteractiveCalculatorProps } from "@/types/calculators";
+import MathInput from "@/components/form/Input/patterns/MathInput";
+import Equation from "@/components/atomic/Equation";
+import Calculator from "@/components/calculators/lib";
+import * as Styled from "./styles";
+
+const InteractiveCalculator: FunctionComponent<InteractiveCalculatorProps> = ({
```
alexgoff, line 10: instead of a map of custom components, make this a single generic component to bring calculators closer to a strategy pattern

components/calculators/lib/equations.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  return { constants: { A, B, C, D } };
+};
+
+const Equations: Record<Equation, EquationComposer> = {
+  peakAbsoluteMagnitude,
+  supernovaDistance,
+};
+
+export default Equations;
```
alexgoff, line 42: this is the math for the calculators, which also returns the constants used so the LaTeX composer can use them.

components/calculators/lib/fields.ts
```diff
@@ -0,0 +1,22 @@
...
+    },
+    peakAbsoluteMagnitude: {
+      key: "peakAbsoluteMagnitude",
+      precision: 1,
+      label: `${placeholders.peakAbsoluteMagnitude} =`,
+    },
+  },
+};
+
+export default FormFields;
```
alexgoff, line 22: define the form fields for each calculator via props

components/calculators/lib/index.ts
```diff
@@ -0,0 +1,39 @@
...
+      {
+        ...cleaned,
+        result,
+      },
+      locale
+    ),
+  };
+};
+
+export default Calculator;
```
alexgoff, line 39: this is the main Calculator method that pulls together form fields, latex, and calculated result for an equation

components/calculators/lib/latex.ts
```diff
@@ -0,0 +1,76 @@
...
+    A
+  )}\\right )${B}^{\\frac{${displayPeakApparent} - ${displayPeakAbsolute}}{${C}} + ${D}}`;
+};
+
+const LaTeX = {
+  peakAbsoluteMagnitude,
+  supernovaDistance,
+};
+
+export default LaTeX;
```
alexgoff, line 76: the LaTeX composers are a little meatier since they have to do localization and also figure out placeholders for missing values

components/form/Input/patterns/MathInput/index.tsx
```diff
@@ -21,7 +21,7 @@ const MathInput: FunctionComponent<MathInputProps> = ({
       {...props}
       type="number"
       placeholder={placeholder}
-      value={value?.toString() || ""}
+      value={value?.toString() ?? ""}
```
alexgoff, line 24: use `??` instead of `||` because it only responds to `null` or `undefined` and not falsey values like `0`

public/localeStrings/en/translation.json
```diff
@@ -307,11 +307,9 @@
     }
   },
   "calculators": {
-    "peak_absolute_magnitude": {
-      "placeholder": "{{delta}}m15"
-    },
-    "supernova_distance": {
-      "result": "{{distance}} Mly"
+    "output": {
```
alexgoff, line 310: live region text

theme/styles/components/_math.scss
```diff
@@ -1,19 +1,10 @@
-"use client";
-import styled from "styled-components";
-
-export const Output = styled.output`
-  cursor: default;
+.calc-output {
```
alexgoff, line 1: quick CSS class that Temml can stick onto the output

blnkt (approved):

PR #158, opened 2024-04-05
[C] use questions instead of datasets

alexgoff:
JIRA: https://jira.lsstcorp.org/browse/EPO-9129

## What this change does ##

Changes the Supernova Distance & Distribution widget to use the result of questions instead of datasets.

## Notes for reviewers ##

Change is client + API (also EPO-9129) and limited mostly to the Supernova Distance & Distribution content block, plus the dynamic component that moshes all the data together. There are also some new helpers to deal with the sea of typechecking from GQL.

- 3

## Testing ##

- In a Supernova Distance & Distribution widget, add an Interactive Light Curve Tool question under the new "Supernovae from questions field"
- In the client, _if the question has been completed_, observe that a user supernova datapoint is now visible and appears and disappears based on the distance filters. 

## Checklist ##


components/content-blocks/SupernovaDistanceDistribution/index.tsx
```diff
@@ -64,7 +55,8 @@ const SupernovaDistanceDistributionBlock: FunctionComponent<
       <SupernovaDistanceDistribution
         album={imageAlbum}
         instructions={instructions || undefined}
-        {...{ supernovaData, step, userData }}
+        questionData={data}
+        {...{ supernovaData, step }}
```
alexgoff, line 59: since this work requires getting some answer data which has to be done on the client, the data is passed down and deferred to the dynamic component

components/dynamic/SupernovaDistanceDistribution/index.tsx
```diff
@@ -18,41 +23,118 @@ const SupernovaDistanceDistribution = dynamic(
...
   const { t } = useTranslation();
+
   const histogramData = supernovaData.map((value, i) => {
     return {
       bin: i * step,
       value,
     };
   });
 
+  const userData = [...questionEntries]
```
alexgoff, line 78: this is the real complex bit, we have to drill down from question -> widgets attached to the question -> the widget we want specifically -> the datasets attached to that widget -> try to match a dataset to one that was saved while working on the question.

I finally came up with some helpers (`isEntryValid`) to do the repetitive checks of null, correct `__typename`, etc. but it is still very tedious. Any optimizations to this are very welcome.

components/dynamic/SupernovaDistanceDistribution/index.tsx
```diff
@@ -18,41 +23,118 @@ const SupernovaDistanceDistribution = dynamic(
...
+        return (data as WidgetInput).datasetId === source.id;
+      });
+
+      if (!isEntryValid(source, "datasets_supernovaGalaxyObservations_Entry"))
+        return undefined;
+
+      const { title: datasetName, distance, lat, long } = source;
+
+      return {
+        id: `${displayName || title}: ${datasetName}`,
```
alexgoff, line 122: This will show "Supernova A: whateverthenameis"

components/dynamic/SupernovaDistanceDistribution/index.tsx
```diff
@@ -66,26 +148,23 @@ const SupernovaDistanceDistributionContainer: FunctionComponent<
...
-  const filteredBinnedImages = binnedImages.filter(
-    (img): img is { width: number; height: number; url: string } =>
-      typeof img !== "undefined"
-  );
+  const filteredBinnedImages = binnedImages.filter(isDefined);
   const imageMismatch = filteredBinnedImages.length !== imageCount;
 
   return (
-    <WidgetContainer
+    <WidgetContainerWithModal
```
alexgoff, line 155: cleanup to not use the HOC

components/dynamic/SupernovaDistanceDistribution/index.tsx
blnkt: I think technically you could do something more concise by making more assumptions about what type something is, and concerns itself more with drilling down for the fields it wants and only covering cases where it does get back the values it needs, and end up with something like (but probably not exactly like): 
``` 
const userData = [...questionEntries]
    .map((entry) => {
      if (!isEntryValid(entry, "questions_default_Entry")) return undefined;

      const { questionWidgetsBlock, id } = entry;
      const { data } = answers[id] || {};

      if (isUndefined(data)) return undefined;

      const { dataset, title, displayName } =
        questionWidgetsBlock?.[0]?.lightCurveTool?.[0];

      const source = dataset.find((source) => {
        const { distance, lat, long, id } = source;
        if (!distance || !lat || !long) return false;
        return (data as WidgetInput).datasetId === id;
      });

      const { title: datasetName, distance, lat, long } = source;

      if (!distance || !lat || !long) return undefined;

      return {
        id: `${displayName || title}: ${datasetName}`,
        distance,
        lat,
        long,
        magnitude: 0,
      };
    })
    .filter(exists);
```

but I think what you have is more clear than something like this, and would be more easily extendable in the future.  'cause you just know that the next ask is going to be to mix and match more answer outputs with more other widgets.  And with what you have you could introduce a sense of like arrays of whitelisted types in place of the explicit types you match against now.  So tedious, yes, but also thorough and forward-thinking.  Keep on keeping on.

PR #161, opened 2024-04-10
[F] Add number question + validation

### LINKED ISSUE #159: Add validated numeric question type
**Is your feature request related to a problem or planned work?**
Planned work in support of Exploding Stars investigation

**Describe the solution you'd like and scope of work**
Add a new question type designed for numeric inputs. Visually identical to the existing text input but pulling up the numeric keyboard and only allowing numeric input.

Should include responsive validations that can be defined in Craft, allowing logical comparisons (if x > 5 then "a message")

**Additional context**

alexgoff:
## What this change does ##

Adds a new question type for numbers, and an answer validation setup for number questions that can be extended to others.

Resolves #159 

## Notes for reviewers ##

Client + API branch

## Testing ##

- Create a new number type question in Craft, the first validation fields (minimum, maximum, precision) are attached to the `input` element
- Add validator fields, an example is shown below. For very large numbers abbreviations can be used, the comparison value will be parsed as LaTeX
image
- Add question to a page
- View the question in the client, begin input. Validations will be run on blur event so as not to barrage users with notifications on every change.

## Checklist ##


app/[locale]/[investigation]/[page]/layout.tsx
```diff
@@ -73,6 +74,7 @@ const InvestigationPageLayout: (
         {reference}
       </Styled.Main>
       <Pager enableAll={user?.group === "educators"} />
+      <Toaster />
     </ProgressProvider>
   );
 };
```
Comment (alexgoff) [inline app/[locale]/[investigation]/[page]/layout.tsx:80]: move the main toaster up to the layout so validations and other areas have access to it. Widget container modals still have their own toaster so that it is visible in fullscreen

components/form/Input/patterns/Question/index.tsx
```diff
@@ -0,0 +1,3 @@
+import { QuestionInput } from "./styles";
```
alexgoff, line 1: extract `QuestionInput` as a reusable input pattern

components/questions/Number/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    data
+  );
+  const { answer, onChangeCallback } = useAnswer<NumberInput>(id);
+
+  const step = 10 ** -parseFloat(precision);
+
+  const handleChange: FocusEventHandler<HTMLInputElement> = async (event) => {
+    const value = parseFloat(event.target.value);
+
+    if (value !== answer) {
```
alexgoff, line 49: maybe some other questions would benefit from this later so as to reduce unnecessary re-renders.

components/questions/Number/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+      if (validation && validation.length > 0) {
```
alexgoff, line 50: we do a simple check here to see if a question has validations before running the validator.

An alternate would be querying the set of validations and sending that array to the server action, that has the downside though of exposing every validation to the client if a user was to go looking for them.

components/questions/actions/validateQuestion.ts
```diff
@@ -0,0 +1,78 @@
...
+const operators: Record<string, string> = {
+  greaterThan: ">",
+  greaterThanOrEqual: "\\geq",
+  lessThan: "<",
+  lessThanOrEqual: "\\leq",
+  equals: "=",
+  notEqual: "\\neq",
+};
+
+const validateQuestion = async (
```
alexgoff, line 36: all the big work is done in this server action, this keeps the ComputeEngine code off the client. Latency during testing was pretty good but is worth a check on a deployed environment to see how it handles communicating between client, Next, and Craft

PR #162, opened 2024-04-11
[F] Add number question + validation 

### LINKED ISSUE #159: Add validated numeric question type

alexgoff:
## What this change does ##

Adds a new question type for numbers, and an answer validation setup for number questions that can be extended to others.

Resolves #159 

## Notes for reviewers ##

Client + API branch

## Testing ##

- Create a new number type question in Craft, the first validation fields (minimum, maximum, precision) are attached to the `input` element
- Add validator fields, an example is shown below. For very large numbers abbreviations can be used, the comparison value will be parsed as LaTeX
image
- Add question to a page
- View the question in the client, begin input. Validations will be run on blur event so as not to barrage users with notifications on every change.

## Checklist ##


blnkt (approved):
This is so wild and cool and totally working.  This feels very much like a now that we have it the Edu team will think of new ways to use it kind of thing, which is a good problem to have.

PR #165, opened 2024-04-15
[C] allow number, text, and textarea in table cell

alexgoff:
## What this change does ##

Allows standalone text, textarea, and number type questions to be added to a display table

## Notes for reviewers ##

Includes client and API branch.

- 4

## Testing ##

- Create a mix of text, textarea, number, and table type questions
- Create a display table (within the investigation pages, not as a question)
- Create a row of headers to label the questions
- Create a row and add your previously created questions under "previous question"
    - the order added is the order the questions will be displayed

See that:
- The results of the questions are rendered in a read-only format with previously entered answers retrieved
- A mix of questions including mixing standalone questions and table type questions co-exist on single row

## Checklist ##


components/layout/Table/helpers.tsx
```diff
@@ -45,22 +52,48 @@ const buildCell = (
...
   readOnly = false
 ): TableCell => {
-  const { id, questionType, options } = cell;
-
-  if (questionType) {
-    const Question = questionType && QuestionsMap[questionType];
+  const { id, answerType, options } = cell;
+
+  if (answerType && id) {
+    if (readOnly) {
```
alexgoff, line 58: build cells into `readOnly` out here instead of having the question cells figure out if they are read-only or not

components/layout/Table/helpers.tsx
```diff
@@ -45,22 +52,48 @@ const buildCell = (
...
+        id: id || undefined,
+        children: <Question {...{ id, questionId, options }} />,
+      };
+    }
   }
 
   return buildStaticCell(cell);
 };
 
+export const buildPreviousQuestions = (questions = []) => {
```
alexgoff, line 78: new helper that builds all the different types of previous questions that could be in a row. Mostly this is used for differentiating tabular questions, which will contain sub-questions, from "primitive" question types

components/questions/TabularQuestion/ReadOnly/index.tsx
```diff
@@ -1,8 +1,47 @@
...
+import { Option } from "@/components/shapes/option";
+import { getLabelByValue } from "@/components/questions/utils";
+
+interface ReadOnlyCellProps {
+  questionId: string;
+  id: string;
+  options?: Array<Option>;
+}
+
+const ReadOnlyCell: FunctionComponent<ReadOnlyCellProps> = ({
```
alexgoff, line 22: move the work of figuring out how to display the answer in a read-only state out to the readonly cell

blnkt (changes requested):
In a static table, when the row header is used in combination with (i.e. in the same row as) previous questions cells, the row header is amended to the row instead of prepended to the row (i.e. is always last in the row instead of always first). See screenshots: in the frontend screenshot, the first row after the table header is what corresponds to the CMS screenshot


Screen Shot 2024-04-18 at 11 31 01 AM
Screen Shot 2024-04-18 at 11 30 02 AM

PR #167, opened 2024-04-16
[C] source selector fixes

alexgoff:
## What this change does ##

Fixes for Source Selector

- Use preview URL from Canto to get consistent 1000px image
- Display controls always (no more difference in fullscreen and on-page display)
- Unwrap and re-map legacy format data
- Create error boundary around widgets to prevent full page crashes

## Notes for reviewers ##

Client only

## Testing ##

- Inspect in console to see that images in the source selector are fixed at 1000px
- Verify that controls are visible and work
- Upload a legacy format light curve file to a dataset, attach to a widget, and verify the alerts show.

[ZTF19aavjzrz_r_lightcurve.json](https://github.com/lsst-epo/investigations-client/files/15003462/ZTF19aavjzrz_r_lightcurve.json)


components/questions/Widget/index.tsx
```diff
@@ -77,11 +78,13 @@ const WidgetQuestion: FunctionComponent<WidgetProps> = ({ data }) => {
       <Styled.QuestionLabel
         dangerouslySetInnerHTML={{ __html: instructions || "" }}
       />
-      <Widget
-        data={widget}
-        instructions={instructions || ""}
-        {...{ value, onChangeCallback }}
-      />
+      <ErrorBoundary>
```
alexgoff, line 81: create an error boundary around widgets so that if they are given incorrect data or encounter another issue, we can see an error message without crashing the entire application.

lib/api/hooks/useAlerts.ts
```diff
@@ -1,18 +1,64 @@
...
+  image_id: number;
+}
+
+export interface LegacySource {
+  name: string;
+  band: string;
+  alerts: Array<LegacyAlert>;
+}
+
+const fetchAndParseAlerts = async (url: string): Promise<Array<Alert>> => {
```
alexgoff, line 26: try to convert any legacy format alerts to the format used by the source selector, the only real difference is the source selector expects an array instead of an object, and that it uses `id` instead of `alert_id`

Issue #236, opened 2024-07-19
Fix name of Docker image in `prod` deployment


ericdrosas87:
When I originally stood up the `investigations-client` repo I accidentally left in the name `rubin-obs-client` for the `prod` workflow, fixing that now.

PR #237, opened 2024-07-19
ci: fix docker image

ericdrosas87:


ericdrosas87:
Resolves #237

ericdrosas87:
Resolves #236

PR #271, opened 2024-09-11
fix: tokens refresh on save

alexgoff:
## What this change does ##

There were several issues with the existing token rotation:
- refreshes were performed during server component rendering, which does not set cookies for the client because [streaming has already begun](https://nextjs.org/docs/app/api-reference/functions/cookies#caveats)
- with tokens not passed on as cookies, a new JWT and refresh token would exist in memory only and quickly be lost, the next refresh would use an invalidated refresh token
- no check or re-check for validity before trying authorized mutations

This is a partial "get-it-working-and-no-more" fix to simply implement token rotation that works when needed. There is additional work to be done, at some point:

- user ID is saved as an unencrypted token and used as the source of identity when performing queries and mutations. Editing the cookie would allow getting/setting other user's answers
- the query client should receive the same authexchange work
- no clean way of getting the user without multiple methods, user is retrieved only from JWT and not validated against the viewer queries.

## Notes for reviewers ##

Frontend only, will add a backend change at a later point to update the token times.

## Testing ##

- before doing anything, you may want to update the JWT settings in the API or project.yaml to be <15min so you aren't waiting around a lot
- log in
- do some answers
- wait around til the JWT is expired
- look at your `craftRefreshToken` cookie before saving 
- save, see that it was successful
- see that the `craftRefreshToken` has changed


components/auth/dialogs/Activate/actions.ts
```diff
@@ -26,7 +26,7 @@ export async function activate(code: string, id: string) {
 
     if (token) {
       // refresh cookies to get the updated user status
-      await refreshToken(token);
+      await refreshTokens(token);
```
alexgoff, line 29: this should be replaced by a proper viewer query instead of refreshing the tokens, but alas

components/auth/serverHelpers.ts
```diff
@@ -34,88 +33,31 @@ export function setAuthCookies(data: AuthFragmentFragment) {
...
-        craftUserId = refreshedData.craftUserId;
-      }
-    } catch (error) {
-      craftToken = undefined;
-      craftRefreshToken = undefined;
-      craftUserStatus = undefined;
-      craftUserId = undefined;
-    }
-  }
+export const getAuthCookies = cache(() => {
```
alexgoff, line 48: do not try to perform refresh when retrieving the cookies, also use `cache` so server requests get a single result

components/auth/serverHelpers.ts
```diff
@@ -16,11 +15,11 @@ export const COOKIE_OPTIONS = {
 } as const;
 
 export function setAuthCookies(data: AuthFragmentFragment) {
-  const { jwt, jwtExpiresAt, refreshToken, refreshTokenExpiresAt } = data;
+  const { jwt, refreshToken, refreshTokenExpiresAt } = data;
 
   cookies().set("craftToken", jwt, {
     ...COOKIE_OPTIONS,
-    expires: jwtExpiresAt,
+    expires: refreshTokenExpiresAt,
```
alexgoff, line 22: this was a big source of issue, the JWT cookie was deleting when the JWT expired. It's fine for the JWT to expire and still read information out of it, if it's deleted while there's still a valid refresh token then the user has been logged out needlessly

lib/fetch/mutate.ts
```diff
@@ -0,0 +1,89 @@
...
+interface MutationClientProps {
+  skipExpirationCheck?: boolean;
+  skipAuthorizationHeaders?: boolean;
+}
+
+type MutationClientFactory = (
+  options?: MutationClientProps
+) => Client["mutation"];
+
+const mutationClient: MutationClientFactory = (options = {}) => {
```
alexgoff, line 32: created a new mutations client with an auth exchange set up that adds the JWT as an authorization header and handles both pro-active and reactive expired auth checks. Only implemented in save answers, trying to contain this change

ericdrosas87 (approved):

PR #380, opened 2026-02-17
clean up output of bucket name and subfolder

jrosati-burwood:
clean up output of bucket name and subfolder


.github/workflows/build-and-push.yaml
gnublet: Can you add `echo date=$(date +'%Y%m%d') >> $GITHUB_OUTPUT`? This is to populate the above date output whiich is required for `bucket_subfolder`

PR #401, opened 2026-04-26
feat: added `SwappableOrbits` configuration

ericdrosas87:
This PR adds the following

- Swappable `OrbitalSim` configuration
- Reverted generated fragment doc because it was putting the chicken before the egg with the codegen
- Moving Source Selector

The immediate use case for these updates is the Hazardous Asteroids port.


ericdrosas87:
FYI @jeffinnes @soshwosh

PR #415, opened 2026-05-06
feat: add a new OrbitalSimReview component for display on the review page

### LINKED ISSUE #412: Add `OrbitalSim` to question review components
An `OrbitalSim` sub-component needs to be added to `/components/questions/Review/Widget/` so that selected observations render as expected on the review page. Currently the `OrbitalSim` questions/answers aren't be shown at all.

jeffinnes:
## What this change does ##

Resolves #412 

This PR adds a new `OrbitalSimReview` component and registers it in the `WidgetReviewWrapper` component. This change will add the OrbitalSim to the review page where appropriate, per the question type.

## Notes for reviewers ##

This new component is very similar to the existing `OrbitalSimQuestion` component but diverges in two notable ways.

1. It does not include the `SelectionList` component
2. The props passed into the context provider for things like `showTimeControls` and `allowOrbitRotation` are hardcoded to `false`
3. `updateSelectedAnswer` is passed a no-operation function to keep the types happy and ensure the selected observation cannot be updated on the review page.

Creating a dedicated "review" version of the component and omitting the `SelectionList` component follows the existing patterns so I felt it justified not refactoring the "question" version of the component to accommodate the "review" versions needs.

## Testing ##

To test this change:

1. Make sure you have an OrbitalSim with interactable observations in your investigation.
2. Select and observation in the OrbitalSim
4. Navigate through the rest of the investigation or us the slide out menu to jump to the review page (`Quick Access > Review you answers`)
5. Observe that the OrbitalSim loads with your saved selection marked as the active one active.

## Checklist ##

### DIFF
--- a/components/questions/Review/Widget/OrbitalSim/index.tsx
+++ b/components/questions/Review/Widget/OrbitalSim/index.tsx
@@ -1,0 +1,62 @@
+"use client";
+import type { FunctionComponent } from "react";
+import type { WidgetReviewProps } from "..";
+import { useState, useEffect } from "react";
+import {
+  OrbitalSimProvider,
+  OrbitalSim,
+} from "@rubin-epo/epo-widget-lib/OrbitalSim";
+import * as OrbitalSimStyled from "@/components/content-blocks/OrbitalSim/styles";
+
+const OrbitalSimReview: FunctionComponent<WidgetReviewProps> = ({
+  data,
+  value,
+}) => {
+  const [dataObj, setDataObj] = useState(null);
+
+  const url =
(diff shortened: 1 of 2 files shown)


ericdrosas87:
note: this looks good, at first I was thrown off because the selected answer isn't shown, but I verified that in `Exploding Stars` the `SourceSelector` answer isn't shown either. I'll approve and we'll get feedback about what the ed group thinks once this is in `dev`.

jeffinnes:
Yeah at first I had included it, but when I went to see how the other investigations handled the reset button I found that they had just omitted the `SelectionList`.

ericdrosas87 (approved):

Issue #413, opened 2026-05-01
Add `Asteroid Size` equation


ericdrosas87:
To support the new `Asteroid Size` question from Craft, we need to add a new equation function. This new equation can be based on [distance modulus](https://github.com/lsst-epo/investigations-client/blob/main/components/calculators/lib/equations.ts#L15), but rather than taking `M` and `m`, they should take `H` and `p` fields. Ardis has more details on this.

PR #417, opened 2026-05-07
Er add asteroid size equation

ericdrosas87:
Adds a new equation to the client and its corresponding formula. This is reliant on a Craft project config update already in the `develop` branch of `investigations-api`.

Note the following:
- **components/calculators/lib/config.ts**
  - Where the parameters for the equation and formula are defined, including the LaTeX that live-updates on the page, specifying the inputs automatically creates the input fields in the client
- **components/calculators/lib/equations.ts**
  - Where the actual mathematical calculations and rounding occur
- **types/calculators.d.ts**
  - Where the type for the formula is defined

The actual JSX/formatting occur in the following places:

- **components/calculators/lib/index.ts*
  - What actually builds the calculator widgets
- **components/calculators/Static/index.tsx**
  - For statically rendering the defaults for a formula or values inputted on a previous page
- **components/calculators/Interactive/index.tsx**
  - For interactive (questions) versions of the equation

### DIFF
--- a/components/calculators/lib/config.ts
+++ b/components/calculators/lib/config.ts
@@ -18,6 +18,21 @@ export const Variables: Record<string, Variable> = {
     precision: 1,
     placeholder: "M",
   },
+  albedo: {
+    key: "albedo",
+    precision: 2,
+    placeholder: "p"
+  },
+  absoluteMagnitude: {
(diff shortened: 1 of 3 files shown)


ericdrosas87:
resolves #413

PR #418, opened 2026-05-08
fix: update `SourceSelector` review widget to handle the `MovingSourceSelector` widget

### LINKED ISSUE #416: Fix `MovingSourceSelector` review component so that only one circle shows
The review component for the `SourceSelector` component needs to be updated to correctly handle the multiple sources (circles) that are present for a moving source. Only the first image and the first source need to be visible for the moving source configuration of the `SourceSelector`.

jeffinnes:
## What this change does ##

resolves #416 

This PR updates the `SourceSelectorReview` component to handle cases where the `sourceSelector` data object has the `hasMovingSource` property set to `true` (indicating this was a moving source selection question).

It does so by updating the logic to conditionally handle data and render the proper component in the same way the question version of the widget does.

## Notes for reviewers ##

This fix depends on [this related fix PR](https://github.com/lsst-epo/epo-react-lib/pull/425) in the `epo-react-lib` being merged and deployed. If that has not been done prior to review testing you will need to `yarn link` the `epo-widget-lib` package on branch `ji-424-update_MovingSourceSelector_widget_...` so the `isDisplayOnly` can be handled correctly in the `MovingSourceSelector` component.

## Testing ##

I tested this change in my local `investigations-client` on the HazAst review page and verified that it doesn't break the existing non-moving source selector in the Exploding Starts review page.

## Checklist ##

### Before:

image

### After:

image


jeffinnes:
> This fix depends on https://github.com/lsst-epo/epo-react-lib/pull/425 in the epo-react-lib being merged and deployed. If that has not been done prior to review testing you will need to yarn link the epo-widget-lib package on branch ji-424-update_MovingSourceSelector_widget_... so the isDisplayOnly can be handled correctly in the MovingSourceSelector component.

This also means I'll need to add one more commit, to this PR, bumping the version of the `epo-widget-lib` package after the referenced fix is deployed.

components/questions/Review/Widget/SourceSelector/index.tsx
```diff
@@ -16,20 +20,40 @@ const SourceSelectorReview: FunctionComponent<
   const [{ sources, json, imageAlbum }] = dataset;
   const { selectedSource = [] } = value || {};
 
+  const hasMovingSource = sourceSelector[0]?.hasMovingSource ?? false;
+
   const { data: alertData = [], isLoading } = useAlerts(json[0].url);
 
-  const { alerts, size } = combineAlertsAndImages(alertData, imageAlbum || []);
+  /* const { alerts, size } = combineAlertsAndImages(alertData, imageAlbum || []); */
```
soshwosh: **💣 chore:** If you don't need this commented-out code and the one on line 32 anymore, you can remove them before merging. If we do need them for something in the future, feel free to add a TODO!‏

soshwosh (approved):

components/questions/Review/Widget/SourceSelector/index.tsx
jeffinnes: Oh good catch! I did forget to remove those before opening the PR.

PR #423, opened 2026-05-11
fix: fixed OrbitalSim modal and fullscreen styling

### LINKED ISSUE #408: Implement `WidgetContainerWithModal` component for `OrbitalSim`
Currently, the `WidgetContainerWithModal` will not render as expected for the `OrbitalSim`. It will shrink down the size of the `OrbitalSim` to a small square in the middle of the page. The full-screen mode also does increase the size of the canvas.

The `WidgetContainerWithModal` or its styling need to be updated so the `OrbitalSim` widget styling does not shrink, and the full-screen mode shows an increased size of the canvas.

soshwosh:
Resolves #408 

## What this change does ##

The `flex-direction:column` and `align-items: center` rules on `WidgetBlock` was overriding the `width:100%` rule coming from the `WidgetContainer` that wrapped OrbitalSim. This change makes some styling adjustments to allow the OrbitalSim to take up more width and to go properly fullscreen. Additionally, some centering for other widgets on smaller dimensions was added.

## Testing ##

1. In your Investigations Craft instance, add a Question Block to any page in an investigation with an OrbitalSim widget entry. 
2. Additionally, add other entries of other widget types for comparison.
3. In the client, navigate to the page with the widget question entries. 
4. Ensure the `OrbitalSim` and other widgets take up a majority of the width of the page and are not taking a narrow width.
5. Click the fullscreen button on each widget and confirm the dimensions of the widgets increase to fill the screen dimensions without altering their aspect ratios drastically enough to affect the widget contents.

## Checklist ##

### Before:
image
image

### After:
image
image

### DIFF
--- a/components/layout/WidgetContainer/styles.ts
+++ b/components/layout/WidgetContainer/styles.ts
@@ -41,6 +41,10 @@ export const WidgetContainer = styled.div`
   max-width: 100%;
   max-height: 100%;
   color: var(--widget-text-color, initial);
+
+  @container (min-width: 800px) {
+    width: 90%;
+  }
 `;
 
 export const WidgetHeader = styled.header`
@@ -69,6 +73,7 @@ export const WidgetBody = styled.div`
 
 export const WidgetRow = styled.div`
   display: flex;
+  justify-content: center;
   flex-direction: column;
   gap: var(--widget-padding);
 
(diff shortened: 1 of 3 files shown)


components/content-blocks/OrbitalSim/styles.ts
```diff
@@ -3,6 +3,7 @@ import styled from "styled-components";
 export const OrbitalSimWidgetContainer = styled.div`
     min-height: 500px;
     width: 100%;
+  }
```
jeffinnes: **❔ question:** ‏Is this closing curly brace something styled components needs? If not, could this cause issues by prematurely closing a CSS block after the styled component code is processed into CSS?
(These aren't leading questions, I genuinely don't know. 😅)

soshwosh: That's definitely just left over from when I was trying some container queries on that styled component, good catch!

jeffinnes (approved):

PR #428, opened 2026-05-14
fix: improve width styles for the `OrbitalSim` `SelectionList` components in the two-column layouts

### LINKED ISSUE #426: Fix `SelectionList` styling in two-column layout `OrbitalSim`
The `SelectionList` component styling should be updated so it is allowed to take up more room when the `OrbitalSim` is presented in a two-column layout.

`SelectionList` component indicated by the red box
Image

The current `max-width` is set in `components/content-blocks/OrbitalSim/styles.ts`.

jeffinnes:
## What this change does ##

Resolves #426 

This PR updates the `max-width` CSS property for the `SelectionListWrapper` to display in a better way in the two-column layouts. It also fixes a bug where the flexbox rules were causing the `WidgetContainer` to become very narrow when the parent container was smaller than 800px. (We did not anticipate the widget being used in the two-column layout when issue #408 was addressed.)

It also moves the selection list below the `OrbitalSim` to match the pattern of other widgets across the investigations. This has been discussed in Slack and was Jose's recommendation.

## Testing ##

To test this change, add an `OrbitalSim` question block with selectable observations to a two-column layout container.

## Checklist ##

### Before:
image

### After:
image

### DIFF
--- a/components/content-blocks/OrbitalSim/styles.ts
+++ b/components/content-blocks/OrbitalSim/styles.ts
@@ -6,10 +6,14 @@ export const OrbitalSimWidgetContainer = styled.div`
 `;
 
 export const SelectionListWrapper = styled.div`
-    max-width: 30%;
+    max-width: 40%;
     margin-top: var(--PADDING_SMALL);
     margin-bottom: var(--PADDING_SMALL);
     button {
         padding: 0px;
+    }
+
+    @container (width < 800px) {
+        max-width: 100%;
     }
 `;
--- a/components/layout/WidgetContainer/styles.ts
+++ b/components/layout/WidgetContainer/styles.ts
@@ -45,6 +45,10 @@ export const WidgetContainer = styled.div`
   @container (min-width: 800px) {
     width: 90%;
   }
+
+  @container (width < 800px) {
+    width: 100%;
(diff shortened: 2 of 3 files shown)


jeffinnes:
@ericdrosas87 just fyi.

soshwosh (approved):
**🔨 issue:** ‏It seems there's some hover styling applied to some empty buttons rendered under the one real button. But this may be a separate bug from the one this ticket resolves and may warrant a new ticket. Ignore this comment if so!

‏‏

Image

Image

jeffinnes:
> **🔨 issue:** ‏It seems there's some hover styling applied to some empty buttons rendered under the one real button. But this may be a separate bug from the one this ticket resolves and may warrant a new ticket. Ignore this comment if so!

I have confirmed that this is technically an issue of the data vs. widget configuration. When the "Contains Swappable Orbits" setting is enabled but the dataset has potential orbits instead of swappable orbits, the widget adds a button for every one of those potential orbits. The buttons appear empty because the potential orbits don't contain the right data.

I don't think this is an issue we need to solve, but it will be a good thing to keep in mind should the Ed. team ever run into this.

PR #433, opened 2026-05-27
JI 431 update orbital sim with object details to support the randomized 1 of x neos data

### LINKED ISSUE #431: Enable OrbitalSim with object details to support the randomized 1 of X NEO data
The `OrbitalSim` implementation needs to be updated such that when a user first navigates to an investigation page containing an `OrbitalSim` with `Show Details Table` enabled they are randomly assigned 1 of 10 `neos` from the dataset to show in the `OrbitalSim`.

A mechanism needs to be implemented to "save" which `neo` they were assigned so they are shown the same `neo` on subsequent visits.

Finally, the Ed team will need a way to override the randomization for testing each `neo` in the dataset.

jeffinnes:
## What this change does ##

Resolves #431

This PR refactors the `content-blocks` implementation of the `OrbitalSimWidget` to randomly assign a Near Earth Object (NEO) from the dataset to the user when they first visit the page IF the `showDetailsTable` flag is set to true. This assignment is stored using the `StoredAnswerContext` context provider and `useAnswer` hook.

To support functionality for educators (and content managers) to override the assigned NEO and support the use of the context provider, the widget had to be converted to a client component. This is a diversion from the normal pattern for `content-blocks` implementations for widgets, but is required to access the data persistence tooling.

## Notes for reviewers ##

Three things to note:

1. A new `components/auth/serverActions.ts` has been made to strictly force the `isUserAnEducator` server-side helper function to stay server-side. This technically should go in the `components/auth/serverHelpers.ts` file, but the way that was authored, allows the functions defined within to convert to client-side functions if imported into a client component. Their functionality requires the server environment; _however_, you cannot use the `"use server";` declaration in a file that exports JSON objects. Refactoring, this and tracking down any possible issues that result from will need to be addressed in a separate issue because the change could widely affect the entire Investigations site.
2. Only an English translation has been added for the educator-only helper text. A separate issue has been created to get translations for both missing `widgets.orbital_sim` items.
3. There is minor issue that will not impact students and only minimally impact educators. Once the "Save" button writes the assigned index to the database, hard refreshing the page with the OrbitalSim will revert the displayed NEO to the originally assigned/saved NEO, _even if the educator override tool changed the NEO and then the user used the "Save" button again_. I have confirmed that the updated index gets written to the database as expected but the hard reload always sets it back to the first saved value. After extensive troubleshooting and `console.log`ing, the behavior seems to indicated that Next.js is caching the `fetch`, but it doesn't quite match up because if I save the data without an assigned index, and then again after getting an assigned index, that assigned index is retrieved from the database as expected. I also noticed that the save functionality writes all new records on every save instead of updating the existing ones which might be contributing. While troubleshooting, I briefly recreated this issue with one of the text based answers where it reverted to the original value I entered, but I couldn't reliably duplicated it. In the end, students won't actually be impacted by this because they are not able to override their NEO index assignment and educators will only be impacted if they are hard refreshing the investigation.  This probably deserves to be investigated in more depth when we have time just incase it's causing issues for students in other parts of the investigation that haven't been reported to us.

## Testing ##

I tested this change by setting up the four pages that require this functionality locally, testing that they work correctly and testing the other instances of the OrbitalSim to ensure there weren't any regressions with their functionality. I ran through my testing logged in as a student and as an educator.

To fully test this change, you will need to add the OrbitalSim widget to different pages using different datasets that contain an array of NEO's with the new data structure. (object details defined in the NEO object) See the `epo-investigations-dev` channel on Slack or chat Jeff for a link to the Dropbox location that holds the updated data sets.

You will also need to `yarn link` the `epo-widget-lib` on `develop-v3` to pick up the change that supports multiple NEOs and the new data structure. (As of starting this PR, that [change](https://github.com/lsst-epo/epo-react-lib/tree/f40017b3e12b3fff51ac2912fcd4560c1023f30a) hasn't been included in a release)

Finally, you'll need to add your user account to the "educators" group in Craft.

Pro Tip: you can quickly test the change in the student view by using an incognito browser window, not logging in, and navigating directly to your page containing the OrbitalSim.

## Checklist ##


components/content-blocks/OrbitalSim/OrbitalSimWidget.tsx
```diff
@@ -1,76 +1,157 @@
...
+    useAnswer<WidgetInput>(orbitalSimTool[0]?.id || "");
 
-    const dataObj = use(getOrbitalData(url));
+  const assignedNeoIndex = value?.assignedNeoIndex;
+  const addTimeControls = orbitalSimTool[0]?.addTimeControls;
+  const allowOrbitRotation = orbitalSimTool[0]?.allowOrbitRotation;
+  const showDetailsTable = orbitalSimTool[0]?.showDetailsTable;
+  const swappableOrbits = orbitalSimTool[0]?.containsSwappableOrbits;
 
+  if (url === "") {
```
ericdrosas87: nitpick: `if(!url)` should suffice here

components/content-blocks/OrbitalSim/OrbitalSimWidget.tsx
```diff
@@ -1,76 +1,157 @@
...
+      const res = await response.json();
+      setDataObj(res);
     }
-    const dataObj = await response.json();
-    return dataObj;
-}
 
-export default OrbitalSimWidget;
\ No newline at end of file
+    if (dataObj === null) {
```
ericdrosas87: nitpick: Same here, `if(!dataObject)`

components/auth/serverActions.ts
```diff
@@ -0,0 +1,29 @@
+"use server";
+
+/**
+ * This server action should probably live in @/components/auth/serverHelpers
```
ericdrosas87, line 4: thought: let's chat more about this in an A&D

components/content-blocks/OrbitalSim/OrbitalSimWidget.tsx
```diff
@@ -1,76 +1,157 @@
...
+      setIsEducator(await isUserAnEducator());
+    }
+    callServerFunction();
+  }, []);
 
-/**
- * to-do: use `fetcher` from `/lib/api` instead
- */
-async function getOrbitalData(url:string) {
+  useEffect(() => {
```
ericdrosas87, line 87: question: Are you certain these `useEffect` usages are necessary?

jeffinnes: "Certain" is certainly too strong of a word. 😉

For the first useEffect that's calling the `isUserAnEducator`, the alternative I had considered would be calling the `isUserAnEducator` function in the `components/factories/ContentBlockFactory/ContentBlockFactory.tsx` parent factory component and passing the result into the component. But that felt like the wrong call since it would only be getting used for this specific widget.

For the second one that assigns the index it's possible I've gotten stuck in a way of thinking, but my reasoning is that assigning the index is dependent on the `dataObj` being populated and it is always `null` on the first render. So I we it to reevaluate once `dataObj` is populated.

components/content-blocks/OrbitalSim/OrbitalSimWidget.tsx
jeffinnes: Oh yeah, it looks like we might have tied the logic gates too literally to their state initializations! I'll sneak in updating these in the question version of the widget as well.

soshwosh (commented):
I was able to confirm the changes worked as expected, but when regression testing I saw some errors. They are (now) non-blocking, as in I can still use the investigation, but there is one (the Unexpected Token one) that I have that @jeffinnes doesn't see. @ericdrosas87 when you get the chance, can you act as a control and spin up this branch and see what errors, if any, you can see? I'll describe the steps I took that led me to see this error below.
Image

Regression Testing:
1. Confirm that you have pulled this branch, and have no `yarn link`'s in place. At the time of this comment, `epo-widget-lib` has been published to NPM with all the necessary changes.
2. On the first page of Hazardous Asteroids, add an OrbitalSim widget block.
3. Add or create an entry that has the **Contains Swappable Orbits** toggle set to **true**.
4. In this entry, add this dataset: [swappable-dataset.json](https://github.com/user-attachments/files/28371408/swappable-dataset.json)
5. Save the entry and the page. 
6. Navigate to the first page of Hazardous Asteroids in the client.
7. Confirm that the widget loads as expected and you are able to swap between Apollo, Atira and Amor.
8. Note any errors.

ericdrosas87 (approved):

ericdrosas87:
I pointed my local client at `dev` and I didn't see any of the errors you saw Sasha, I did see the usual warnings-as-errors which we should address soon, but I'm approving this and mergin.g
Screenshot 2026-05-29 at 8 44 35 AM

PR #437, opened 2026-06-02
feat: update the orbitalSim widget implementations to retrieve the defaultZoom from Craft

### LINKED ISSUE #435: Modify the Orbital Sim widgets to take the new `defaultZoom` Craft field and pass it as the `defaultZoom` context provider prop
After the `defaultZoom` Craft field has been added to the `OrbitalSim` content block and question blocks, pass it into the corresponding components' context provider as the `defaultZoom` prop.

jeffinnes:
## What this change does ##

Resolves #435 

This PR updates the `content-block` and `question` implementations of the OrbitalSim widget to query the `api` for the `defaultZoom` value that should be passed into the `OrbitalSimProvider`, rather than extracting it from the dataset.

## Notes

My actual modifications to `components/questions/Widget/OrbitalSim/index.tsx` were really just to add the `defaultZoom` logic like in `components/content-blocks/OrbitalSim/OrbitalSimWidget.tsx`, but the code was getting a little unwieldy so I ran the "Format Document" command on it.

## Testing ##
_Note_: To test this change you will need to `yarn link` the `epo-widget-lib` on branch `ji-456-Update_OrbitalSim_widget_to_receive_and_use_a_defaultZoom_prop` 

To test that the **fallback default value** of `0.5` is being applied when a `defaultZoom` value has not been applied to a widget in Craft, navigate to a page displaying that widget and see that it is not zoomed in too close to the sun. It's hard to say exactly because of possible differences in datasets, but for the most part, the `0.5` fallback seems to give a full view of Jupiter's orbit. Prior to this fix, most of my datasets had the default view somewhere around Earth's orbit. (see screenshots below)

To test that the **Craft setting** is being applied as the `defaultZoom`, set the "Default Zoom" value in one of the orbital sim widgets. Setting it to something high like `6.00` and low like `0.01` will demonstrate an easily noticeable change.

_Second Note_: Make sure you save the page in addition to saving your change to the widget. It seems that Craft does not actually save the change to the widget until the page is saved.

## Checklist ##

### Before:
image

### After:
image


soshwosh (approved):

jeffinnes:
I am going to wait to merge this PR until the changes in this [epo-react-lib PR](https://github.com/lsst-epo/epo-react-lib/pull/457) are published so I can update the `epo-widget-lib` dependency version in this codebase.

PR #456, opened 2026-08-03
Ji 453 fix haz ast moving source selector pulling images from netlify

### LINKED ISSUE #453: HazAst moving source selector is pulling images from Netlify instead of Canto
**Describe the bug**
The moving source selector widget on the [detecting-an-asteroid](https://investigations.rubinobservatory.org/hazardous-asteroids/detecting-an-asteroid) page is set to use an album from Canto, however the actual network requests for the images are being made to Netlify. We have received reports that Netlify is blocked by some users schools or organizations.

In addition, we will eventually be shutting down our Netlify presence which will break the widget for all users.

**To Reproduce**
1. Go to the [detecting-an-asteroid](https://investigations.rubinobservatory.org/hazardous-asteroids/detecting-an-asteroid) page
2. Inspect the dev tools `network` tab and note where the images are requested from

**Expected behavior**
The images should be retrieved from Canto

**Screenshots**

Image

**Additional context**
If it turns out we have hard coded Netlify URLs in the app we should check for any other investigations that have had this hard coded and get it fixed now.

jeffinnes:
## What this change does ##
Resolves #453 

This PR fixes the bug where the the MovingSourceSelector widget, used on the 3rd page of the Hazardous Asteroids investigation, was still using the images hosted on Netlify instead of the Canto album that was set in Craft. This was happening because we were passing the `alerts` array, which is parsed from the Alerts Dataset file, directly into the `alerts` prop of the `MovingSourceSelector` component instead of the processed `alertsWithImages` array that has alerts and images paired up.

This was also required in the "review" implementation of the widget.

## Notes ##

**The data set loaded in Craft will need to be updated as part of deploying this fix.**

## Testing ##

To test this fix, replicate the widget setup for the `hazardous-asteroids/detecting-an-asteroid` page in your local investigations-api Craft instance. However, for the Alerts Data set create and upload a .json file with the following data.

_Note: Some of the data points are not directly needed for the widget to work, but they might still be needed by other parties if this data needs to be updated or reviewed later so I am opting to leave it as is._

```
{
    "id": "",
    "name": "hazardous-asteroids",
    "band": "r",
    "dec": 18.74361216,
    "ra": 45.08449499,
    "distance": 141.69708547161875,
    "velocity": 9736.736057325023,
    "redshift": 0.033,
    "sources": null,
    "movingSources": [],
    "alerts": [
        {
            "id": "2003_QZ30_0",
            "error": 0.05,
            "date": 58719.46459,
            "magnitude": 17.58
        },
        {
            "id": "2003_QZ30_1",
            "error": 0.04,
            "date": 58722.4869,
            "magnitude": 17.35
        },
        {
            "id": "2003_QZ30_2",
            "error": 0.03,
            "date": 58725.49949,
            "magnitude": 17.24
        }
    ]
}
```

Verify the following are true:

- [x] selection circle moves with the asteroid
- [x] label is shown correctly
- [x] review page shows the widget correctly
- [x] Dev Tools Network tab confirms that images are not fetched from Netlify 

## Checklist ##

### DIFF
--- a/components/questions/Review/Widget/SourceSelector/index.tsx
+++ b/components/questions/Review/Widget/SourceSelector/index.tsx
@@ -36,7 +36,7 @@ const SourceSelectorReview: FunctionComponent<
     <>
       {hasMovingSource ? (
         <MovingSourceSelector
-          alerts={alertData}
+          alerts={alerts}
           {...{ selectedSource, isLoading }}
           movingSources={percentageMappedSources}
           width={size}
--- a/components/questions/Widget/SourceSelector/index.tsx
+++ b/components/questions/Widget/SourceSelector/index.tsx
@@ -2,7 +2,7 @@ import { FunctionComponent, useState } from "react";
 import { FragmentType, graphql, useFragment } from "@/gql/public-schema";
 import SourceSelector, {
   SelectionList,
-  MovingSourceSelector
+  MovingSourceSelector,
 } from "@rubin-epo/epo-widget-lib/SourceSelector";
 import useAlerts from "@/lib/api/hooks/useAlerts";
 import WidgetContainerWithModal from "@/components/layout/WidgetContainerWithModal";
@@ -12,7 +12,7 @@ import {
   combineAlertsAndImages,
   percentageMapSources,
   percentageMapSourcesForMovingSources,
-  combineAlertsAndImagesForMovingSources
+  combineAlertsAndImagesForMovingSources,
 } from "@/helpers/widgets";
 
 const Fragment = graphql(`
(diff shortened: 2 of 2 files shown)


soshwosh (approved):
In testing this PR, a bug was discovered: When the `radius` field of the sources in the Source Selector > Dataset field is set, the user can click on any source in the widget and get the "Success" message, as well as observe the correct source being circled. When no  `radius` value is provided, the widget works correctly. This can be addressed in a different ticket and PR.

ericdrosas87:
> In testing this PR, a bug was discovered: When the `radius` field of the sources in the Source Selector > Dataset field is set, the user can click on any source in the widget and get the "Success" message, as well as observe the correct source being circled. When no `radius` value is provided, the widget works correctly. This can be addressed in a different ticket and PR.

If I'm remembering correctly, it's not that you can click anywhere when the field is populated, but that the size of the radius is very large in > 1 values. A radius of `0.1` for example would be more precise. This behavior was observed [here](https://rubin-obs.slack.com/archives/C08ALS6NTHV/p1778709704374099), and I thought the ed group had followed-up and said they never use the `radius` field anyway, but that may have been during a call because I don't see it on Slack.

